Issue #4, opened 2020-04-01
Colour best times of the week to sell?


mikebryant:
Or maybe on each individual prediction, whether it's good or bad?

mikebryant:
We're now doing this :)

PR #12, opened 2020-04-06
Add gradients to price cells

jkrems:
Preview:

Screen Shot 2020-04-06 at 9 29 50 AM

Fixes https://github.com/mikebryant/ac-nh-turnip-prices/issues/4

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -52,14 +52,59 @@ $(document).on("input", function() {
     return;
   }
 
+  function rangeToGradient({ min, max }, {weekMin, weekMax}) {
+    const bad_color = '#ffafa9';
+    const good_color = '#cef3a3';
+
+    function scale(value) {
+      if (value === buy_price) return 50;
+      if (value < buy_price) {
+        return 50 - (buy_price - value) / (buy_price - weekMin) * 50;
+      }
+      return 50 + (value - buy_price) / (weekMax - buy_price) * 50;
+    }
+
+    // Two scales with buy pct in the middle.
+    const buy_pct = scale(buy_price); // always at 50%
+    const min_pct = scale(min);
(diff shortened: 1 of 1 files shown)


jkrems:
Thanks a lot for creating this tool, btw! It's super useful. :)

louh:
ooh I was just looking at this issue too.

I'm not involved in this project (but I use this actively and I'm interested in what's going on behind the scenes) and I have to ask, how do you interpret these gradients? Obvious the green means "good" and "red" means "bad" but there's a lot of different scales of each ... and it may be introducing a lot of visual noise.

I don't mean to be overly critical, just my observation (as a UI designer/programmer) - perhaps it might be easier to interpret with solid color backgrounds, using a much more simple color ramp?

For example:
image

Green = sell!
Yellow = hold for something better
Red = it's bad, look at another island perhaps?

jkrems:
I totally agree that having a straight model for clear expected profits and coloring based on that would be better. To me the coloring above expresses "how likely is the price on that day to be loss or profit". A single color, especially for the summary row, would imply that we *know* that it's a good day to sell which given the ranges for the following days is tricky to say.

I had experimented with solid colors for days that are all good or all bad but to me at least that made the graph harder to read. Because all good, in absence of more sophisticated statistical analysis, can mean plenty of different things, depending on the relation to future prices and their ranges.

In summary: I consider my suggested coloring here a stop-gap measure until somebody comes up with a better way to process the data to have clear signals. This PR only renders the data (the price ranges) that already exists.

khoover:
The model's no longer like in NL, where even though ranges might overlap, there is a specific half-day where the price will be the maximum?

mikebryant:
I went with a different direction here, but thanks for your contribution. Hope you keep enjoying the app :)

PR #16, opened 2020-04-06
Probability updates

alexwh:
Calculate and display the probability in the UI
Still needs to recalculate based on removed patterns. e.g. small spike from random is 35%, but if the only patterns in the list are random and small spike, then it should be 35/(35+20), with 20 being random's chance

### DIFF
--- a/index.html
+++ b/index.html
@@ -34,6 +34,16 @@
   </p>
   <form>
     <p>
+      <label>Previous week's pattern. This is used for probabilities only.
+        <select id="previous_pattern">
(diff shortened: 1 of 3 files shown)


alexwh:
Implemented the percentage reevaluation mentioned in the first post, I think the probability maths should be correct.

alexwh:
Phew that was a heck of a merge conflict. Will test shortly to see if everything is on fire.

alexwh:
I believe everything is now merged/bugfixed correctly.

mikebryant:
Thanks for this, I've got a version of this merged in master now :)

PR #18, opened 2020-04-07
Refactored scripts.js

Trevor-Welch:
This commit cleans scripts.js, abstracts functions, and makes it a bit more easier to read - all while changing absolutely nothing about the functionality of the app.

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -1,63 +1,86 @@
-$(document).ready(function () {
-  // load sell_prices from local storage
-  try {
-    const sell_prices = JSON.parse(localStorage.getItem("sell_prices"));
+//Reusable Fields
+const getSellFields = function () {
+  let fields = []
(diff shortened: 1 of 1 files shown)


js/scripts.js
```diff
@@ -1,63 +1,86 @@
...
+  } else {
+    return prices
   }
+}
 
-  localStorage.setItem("sell_prices", JSON.stringify(sell_prices));
+const getSellPrices = function () {
+  //Checks all sell inputs and returns an array with their values
+  return res = sell_inputs.map(function (input) {
+    return input.value || ''
```
mikebryant, line 73: The analysis functions expect the data as numbers, not as strings. I think this is missing the `parseInt` call that was present previously?

mikebryant:
I've fixed up and merged, thanks for the help :)

PR #27, opened 2020-04-07
Add support to remember prices in query string.

peter50216:
This makes linking to a particular price chart easier.

(I use "." as separator instead of "," so the URL looks a little nicer, URLSearchParams seems to escape "," to "%2C")

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -69,7 +69,7 @@ const isEmpty = function (arr) {
 
 const getFirstBuyState = function () {
   return JSON.parse(localStorage.getItem('first_buy'));
-};
+}
 
 const getPricesFromLocalstorage = function () {
   try {
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -78,10 +78,9 @@ $(document).on("input", function() {
     sell_prices.push(parseInt($("#sell_" + i).val()));
   }
 
-  localStorage.setItem("sell_prices", JSON.stringify(sell_prices));
-  const params = {"prices": sell_prices.map((x) => isNaN(x) ? "" : x).join(".")};
-  const query_string = `?${new URLSearchParams(params).toString()}`;
-  window.history.replaceState(null, null, query_string);
+  if (!window.price_from_query) {
+    localStorage.setItem("sell_prices", JSON.stringify(sell_prices));
+  }
 
   const is_empty = sell_prices.every(sell_price => !sell_price);
   if (is_empty) {
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -68,7 +68,7 @@ const isEmpty = function (arr) {
 }
 
 const getFirstBuyState = function () {
-  return JSON.parse(localStorage.getItem('first_buy'));
+  return JSON.parse(localStorage.getItem('first_buy'))
 }
 
 const getPricesFromLocalstorage = function () {
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -8,8 +8,8 @@ function get_prices_from_localstorage() {
   return sell_prices;
 }
 
-function get_prices_from_hash() {
-  const params = new URLSearchParams(window.location.hash.substr(1));
+function get_prices_from_query() {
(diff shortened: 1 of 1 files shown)


mikebryant:
Thought - is there a danger here that someone clicks on a link from someone else, and therefore wipes their own data?

theRTC204:
A query string seems more contextually appropriate for what you're doing here. The hash parameter should be used to point to a resource on a page, while the query string is meant to set values to parameters.

peter50216:
Yeah currently it would override the data in localstorage, I'm not sure how many people actually use this site to record their own data. (I record mine on a spreadsheet elsewhere, so it's handy to have this so I can automatically generate a link to the prediction site :D)

theRTC204:
One possible solution to @mikebryant's comment could be to not write to local storage if there is a URL segment present.

I would imagine that most people use the side to record their own data day to day throughout the week. Everyone that I know who uses it does it that way (and loves that it stores the values for them).

peter50216:
One question is that this currently automatically changes the URL when there's input, so there would always be a URL segment present.

Or should I just not change the URL on input, and explicitly add a "share" button that would generate the URL for copy?

peter50216:
Also if I change this to query string then it can't be automatically changed when input changes (writing to window.location.search reloads the page), but I guess this is not a problem if going for the route in the previous post.

theRTC204:
> Also if I change this to query string then it can't be automatically changed when input changes (writing to window.location.search reloads the page), but I guess this is not a problem if going for the route in the previous post.

What you'd want to do is use the browser's History API to manipulate the history stack (https://developer.mozilla.org/en-US/docs/Web/API/History_API). This is the underlying native API that is used by frontend routers often paired with SPA frameworks like Angular or React.

A call to `history.replaceState()` should do the trick.

theRTC204:
I'm curious, what is the intended value that this change brings? Given sell prices are largely randomized, and very personal to your island, what do you envision using the sharable URL for?

peter50216:
> What you'd want to do is use the browser's History API to manipulate the history stack (https://developer.mozilla.org/en-US/docs/Web/API/History_API). This is the underlying native API that is used by frontend routers often paired with SPA frameworks like Angular or React.
> 
> A call to `history.replaceState()` should do the trick.

Oh yes I forgot about that, thanks for the hint!
Changed to query string.

peter50216:
> I'm curious, what is the intended value that this change brings? Given sell prices are largely randomized, and very personal to your island, what do you envision using the sharable URL for?

The use case I have is that we have a group of friends playing ACNH, and a spreadsheet that everyone use to record their price and see others, so I want a quick link from the spreadsheet to the prediction site to see who got which pattern and still has a chance to have high price.

theRTC204:
It's not my project, so really @mikebryant will have to weigh in his feelings. But my sense is adding a Share button that generates a link which pre-fills the fields seems fine, as long as the URL does not end up writing to Local Storage.

peter50216:
Changed so that this "feature" is hidden from normal user at all, and you have to read the code to know there's a query string that can be used to pre-fill the form.

I feel that if we want to add a "share" button then we might need to have some banner showing that the view is somewhat readonly and you can go back to your save data by clicking some link. I'd leave that to another PR :smile:

Trevor-Welch:
I would be willing to work on a PR for two different views based on URL params, and I would just add a param that is added to the URL from the Share button like ?source='share' or something like that that the page catches and prevents from writing to storage.

mikebryant:
This is looking good, we can definitely add this and iterate :)

Thank you for working on this!

Burry83:
Just bought this to life in a similar way @peter50216 as had a group of mates with a spreadsheet. When linking across thought the previous pattern data is lost and therefore my believe that the % Chance is not calculated. Is there any desire from anyone else to add the previous pattern onto the query string?

Burry83:
I have so much time for this. Just set it all up from our google sheet and it has worked like a charm. Thank you all for this.

PR #30, opened 2020-04-07
Add 'LeastMax' Parameter

jtplatt99:
Add additional logic to compute 'LeastMax', which represents the worst case scenario maximum price achievable during the week (assuming everything is sold on the correct day). I think this is more useful than the current weekMin, since most people won't be selling then, but I'm in favor of displaying both.

I think the parameter is definitely in need of a better name, but it will be helpful in determining a realistic range of prices that the turnips can be sold for to maximize profit.


mikebryant:
I've reworked this a little and merged, thank you!

PR #34, opened 2020-04-08
Front-End Complete Redesign

Trevor-Welch:
This PR absolutely remodels the front-end of this project.

The design is inspired by the Animal Crossing in-game Nook Phone.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -264,8 +264,8 @@ input[type=number] {
 }
 
 .input__radio-buttons input[type="radio"]:checked+label {
-  background: transparent;
-  color: #0AB5CD;
+  background: #0AB5CD;
+  color: #FFF;
 }
 
 .reset-button {
--- a/css/styles.css
+++ b/css/styles.css
@@ -109,11 +109,11 @@ h2 {
 }
 
 .form__row {
-  max-width: 450px;
   display: flex;
   flex-wrap: wrap;
   margin-bottom: 32px;
   justify-content: center;
+  align-items: center;
 }
 
 .form__row h6 {
@@ -137,20 +137,22 @@ h2 {
 .input__group {
   display: flex;
   flex-direction: column;
+  margin: 8px;
+  align-items: center;
 }
 
 .input__group label {
   font-size: 1.2rem;
   font-weight: bold;
   margin-bottom: 8px;
-  opacity: 0.7
+  opacity: 0.7;
+  text-align: center;
 }
 
 .form__flex-wrap .input__group label {
-  text-align: center;
+
   margin-left: 0px;
   margin-bottom: 8px;
-  margin-right: 32px;
 }
 
 .input__form i {
@@ -176,7 +178,6 @@ input {
   font-weight: bold;
   transition: 0.2s all;
   margin: 8px 0px;
-  margin-right: 32px;
 }
 
 input[type=number]:placeholder-shown {
@@ -186,6 +187,12 @@ input[type=number]:placeholder-shown {
 input[type=number]:not(:placeholder-shown) {
   background: transparent;
   color: #0AB5CD;
+}
+
+input[type=number]:placeholder-shown:hover {
+  cursor: pointer;
+  transform: scale(1.1);
+  box-shadow: 0 1px 6px rgba(0, 0, 0, 0.05), 0 3px 6px rgba(0, 0, 0, 0.09);
 }
 
 input[type=number]:focus {
@@ -204,6 +211,16 @@ input[type=number] {
   text-align: center;
 }
 
+input[type=number]:disabled {
+  background: inherit;
+}
+
+input[type=number]:disabled:hover {
+  box-shadow: none;
+  transform: none;
+  cursor: default;
+}
+
 input::-webkit-outer-spin-button,
 input::-webkit-inner-spin-button {
   -webkit-appearance: none;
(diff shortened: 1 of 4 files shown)


promarcel:
I have some comments about the Redesign from UX view. But at first thank you so much for spending much time to create this. My comments are just small details:

* When clicking into an input field the background changes to white with a shadow, this is great but on some displays this does not stand out enough, on Retina it does work, but for example on my BenQs not, its just a small-hard-to-see field. An option could be to try another stronger shadow or color. Another option could be also a light-gray background color.
* The whole UI is large-scale (which is not a criticism, this is fine) but compared to the UI, the output table appears very small. Could we change the size of the text in the table from `0.8rem` to `1rem`?
* Another suggestion for the output table: If there are some lines in the table, it gets a right scrollbar, can we please use JavaScript to dynamically enlarge the whole box so that it doesn't get a right scrollbar?
* Because of the data table we also have a lower scrollbar, on the Mac this is no problem, it looks discreet and doesn't disturb. On other platforms like Windows, a large grey scrollbar appears here. 
Maybe we can adjust the CSS for the browsers that support it, so that a modified light scrollbar is displayed. Example can be  https://css-tricks.com/examples/WebKitScrollbars/

Hope, my comments will help continue improving UI/UX 😃

css/styles.css
```diff
@@ -1,20 +1,345 @@
...
+  border-collapse: collapse;
+}
+
+#turnipTable span {
+  min-width: 80px;
 }
 
 #turnipTable th {
-    text-align: center;
+  min-width: 80px;
```
tylermatchett, line 257: I would remove the min-width and let padding handle it, will help in the vertical space on smaller displays
```suggestion
```

css/styles.css
```diff
@@ -1,20 +1,345 @@
...
+  font-size: 0.8rem;
```
tylermatchett, line 258:
```suggestion
```
This will help with readability

css/styles.css
```diff
@@ -1,20 +1,345 @@
...
+}
+
+
+.table-wrapper {
+  width: 98%;
+  max-height: 50vh;
+  padding: 16px;
+  margin: 0px auto;
+  box-sizing: border-box;
+  overflow-x: scroll;
```
tylermatchett, line 244:
```suggestion
  overflow-x:  auto;
```

The reason for this is that we will set the table width to 100% and this will handle the scroll bar showing up on wider screens that do not need it.

css/styles.css
```diff
@@ -1,20 +1,345 @@
...
+  border-radius: 2px;
+}
+
+#turnipTable {
+  border-collapse: collapse;
+}
+
+#turnipTable span {
+  min-width: 80px;
```
tylermatchett, line 253: Same as above, we should let the padding handle the spacing here.
```suggestion
```

css/styles.css
```diff
@@ -1,20 +1,345 @@
...
+  padding: 16px;
+  margin: 0px auto;
+  box-sizing: border-box;
+  overflow-x: scroll;
+  border-radius: 2px;
+}
+
+#turnipTable {
+  border-collapse: collapse;
+}
```
tylermatchett, line 250:
```suggestion
  width: 100%;
}
```
Allows us to keep some nice spacing on the table on wider displays.

css/styles.css
```diff
@@ -1,20 +1,345 @@
...
+#turnipTable tbody tr {
+  opacity: 0.6;
+}
+
+#turnipTable tbody tr:hover {
+  cursor: default;
+  opacity: 1;
+}
+
+
```
tylermatchett, line 290:
```suggestion
.table-wrapper::-webkit-scrollbar {
  height: 8px;
}

.table-wrapper::-webkit-scrollbar-track {
  height: 8px;
  -webkit-box-shadow: inset 0 0 6px rgba(0,0,0,0.2);
}

.table-wrapper::-webkit-scrollbar-thumb {
  height: 8px;
  background: rgba(0,0,0,0.2);
  -webkit-box-shadow: inset 0 0 6px rgba(0,0,0,0.1);
}
.table-wrapper::-webkit-scrollbar-thumb:window-inactive {
  height: 8px;
  background: rgba(0,0,0,0.2);
}
```

Styles the scroll bar on narrow windows based systems on supported browsers.

tylermatchett, line 290:
```suggestion
#turnipTable .table-pattern {
  white-space: normal;
}

```

This is a style to still allow the pattern to collapse into multiple lines, to aid with the overall width of the table on narrow displays

js/scripts.js
```diff
@@ -89,7 +89,7 @@ const calculateOutput = function (data) {
     var out_line = "<tr><td>" + poss.pattern_description + "</td>"
```
tylermatchett, line 89:
```suggestion
    var out_line = "<tr><td class='table-pattern'>" + poss.pattern_description + "</td>"
```

Allows us to style this element specifically

css/styles.css
theRTC204: Although is does reduce the width of the overall table - the table size is such that the patterns would likely end up with a single word on each line, which greatly reduces readability. IMHO this should stay as `nowrap`.

theRTC204:
image
Contrast on these colors definitely breaks accessibility standards, and even I find it difficult to read.

css/styles.css
tylermatchett: That is a good point, and if there is a scroll bar, I guess it doesn't matter too much if there is a bit extra.

Here are the two views, one with and one without, just to give an idea of how they both look.

With
with wrap

Without
without-wrap

theRTC204: Well, I have another concern with the changes to the pattern names. The patterns are named in such a way that they both match the information extracted from the game code, as well as clearly describe the behavior of the pattern itself.

These pattern description changes do not clearly explain how the patterns will behave. I'm not in favor of the renames.

tylermatchett: I see where you are coming from, I feel like for readability and UX, it doesn't make too much difference as most of the value to a user comes from the prediction data itself and the new names do make it easier to parse. However the loss of matching the game code is a concern.

theRTC204: Perhaps a solution that addresses both needs would be to add a legend on the page that maps the short names to their true names?

tylermatchett: yeah that would work, or a tooltip/hover over effect that has the details, depending on how that would interact with mobile.

index.html
```diff
@@ -58,13 +58,47 @@ <h1>Turnip Prophet</h1>
...
 
     <div class="nook-phone-center">
+
+
+
       <form class="input__form">
+        <div class="form__row">
+          <h6>First-Time Buyer</h6>
+          <div class="input__group">
+            <label>Have you purchased turnips on your own island?<i>(This affects your pattern)</i></label>
```
theRTC204, line 68: This message is not quite right. In the context of the algorithm, we care whether this specific week is the first time they have ever purchased. The way this is worded, it is open to misunderstanding the intention of the feature. This language makes it seem as though the user should choose `yes` every time they buy from their own island.

When the "first buy" setting is enabled, the app will only return predictions that match Pattern 3 and it will return results for every possible buy price regardless of what is entered into the buy price input. We only want this to happen if this is the first time they have ever purchased - for all other weeks, we want to make sure they do not enable this feature.

mikebryant:
This looks really nice. Thank you for doing this!

I've rebased and made some fixes in #50

Issue #13, opened 2020-04-06
Chance of each pattern


louh:
Hi! I love this tool and use it every day with my friends :) It paid off last week when someone with a predicted high hit it (519 bells)!

I've been looking at the reverse engineered code, too, and have an idea I'd like to implement, and I'd be happy to take a stab at a pull request if you're interested. What we know is that each of the four price patterns have a chance of happening depending on the previous week's pattern. (So if you had a decreasing pattern last week, your chance of a large spike this week is 50%). Right now, each of the patterns are shown, but we don't have a sense of what are the odds of each pattern occurring.

I'm thinking of adding a dropdown where you can select which of the four patterns you had the previous week (or "unknown", the default choice). It won't be required, only if you happen to know what it is. The results will then calculate what are the odds of each pattern happening. And when we know some patterns won't happen, the chances will be recalculated based on available patterns. It should give someone much more clarity on what their odds are of hitting a large spike, given their prices and previous week's behavior!

Let me know what you think and if you would like me to help!

mikebryant:
I was looking at this earlier - but I'm totally happy for you to take a stab at this, I love collaboration.

mikebryant:
Here's the WIP I have from earlier today: https://github.com/mikebryant/ac-nh-turnip-prices/tree/random-hacks-towards-probability

mikebryant:
Feel free to take that forward or start afresh :)

alexwh:
I made #16 - seems to work well enough for now. I passed the previous_pattern variable to every function but it might be better to make it global or create a class in the future.

louh:
@alexwh I took my own approach last night, and after reviewing your PR I see we did some similar things and differed in others. Any interest in combining efforts?

mikebryant:
I've now gotten this all merged and integrated - but if there's further changes you think we need, open to more issues or PRs. Thanks for your help!

PR #35, opened 2020-04-08
feat: display %chance of each pattern if known

louh:
This is a PR to address issue #13!

I did refer to the existing branch, but since the codebase has updated pretty significantly over the past day, it was easier to start afresh.

One area that might be up for debate is how to divide up percentages. For me, it's more useful to know the overall percentage of each pattern, rather than the percentage chance of each possibility row. (In other words, I'd prefer to know I have 50% of a large spike, and not 10% of 5 different large spike possibilities.) This might be up for debate?

Let me know what you think, about this or any other part of the PR!

### DIFF
--- a/index.html
+++ b/index.html
@@ -39,6 +39,33 @@
     </div>
     <div class="row">
       <div class="col s12">
+        <h5>Probability</h5>
+        <div class="row">
+          <div class="col s4">
+            <div class="input-field">
+              <select id="previous_pattern">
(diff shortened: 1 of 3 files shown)


mikebryant:
Thanks for this - I rebased and made some changes. Since it's displayed per row I'm strongly in favour of having the probability be for that row. Open to having another display for overall pattern somewhere? Unsure, can see how it looks post-redesign

But very much thanks for your work!

PR #42, opened 2020-04-08
feat: Integrating probability

mikebryant:
Builds off of #16 and #35

I've changed it to calculate probability per-row

### DIFF
--- a/index.html
+++ b/index.html
@@ -39,6 +39,33 @@
     </div>
     <div class="row">
       <div class="col s12">
+        <h5>Probability</h5>
+        <div class="row">
+          <div class="col s4">
+            <div class="input-field">
+              <select id="previous_pattern">
+                <option value="-1" selected>(unknown)</option>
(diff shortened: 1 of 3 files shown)


louh:
Awesome, thank you!

alexwh:
Looks good, better implemented than my first lazy draft haha. About the per row probability - is this the true chance of getting that pattern? I considered calculating it like this but since the pattern is decided at first independantly from the rest of the values, shouldn't they have a chance to appear based on which patterns are remaining, rather than the number of rows remaining? I'm not sure on the stats maths on this, so feel free to correct me.

mikebryant:
I don't believe it's quite right - because I think the actual conditional probability would also be based on where your day value is in the range, and how likely it is a pattern would have matched.

Buuut, I think it's a reasonable approximation. Definitely open to making it more accurate - but a rough guess is likely helpful anyway

PR #51, opened 2020-04-09
Fixing the Reset Button

tylermatchett:
Intercepts the error when the reset can't find the radio button which allows it to continue and actually reset the fields

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -33,7 +33,9 @@ const checkRadioByValue = function (radio_array, value) {
     return;
   }
   value = value.toString();
-  radio_array.find(radio => radio.value == value).checked = true;
+
+  if (typeof radio_array.find(radio => radio.value == value) !== 'undefined')
+    radio_array.find(radio => radio.value == value).checked = true;
 }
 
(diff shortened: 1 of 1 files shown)


mikebryant:
This is actually because I forgot to change the reset code to point at the right button - I've fixed that instead :)

js/scripts.js
```diff
@@ -33,7 +33,9 @@ const checkRadioByValue = function (radio_array, value) {
     return;
   }
   value = value.toString();
-  radio_array.find(radio => radio.value == value).checked = true;
+
+  if (typeof radio_array.find(radio => radio.value == value) !== 'undefined')
+    radio_array.find(radio => radio.value == value).checked = true;
```
theRTC204, line 38: This change does not fix the problem, but just masks it by swallowing the error. The actual issue is caused by an invalid value being passed to `fillFields()` on line 86 of the file.

```js
fillFields([], false, 'unknown')
```
should actually be
```js
fillFields([], false, -1)
```

theRTC204:
RIP my slow review submission, hahaha :(

PR #52, opened 2020-04-09
fix: Reduced overall element size throughout UI

theRTC204:
This change reduces the overall element sizes for the new UI. This was done by reducing paddings and font sizes for all inputs.


mikebryant:
Thanks :)

PR #53, opened 2020-04-09
fix: Improved data table visibility on different devices

theRTC204:
This change introduces new behavior for the data table.

If the device supports a fine pointer (mouse or touchpad) and a screen
width that is less than 1400px, we will set a max height on the table
allowing the horizontal scroll bar to be easily accessable.

If the device supports a coarse input (touch) we will not set a max
height on the table, because horizontal scrolling is a simple swipe.

If the device has a screen that is wider than 1400px we will not set a
max height on the table, as there would not be a horizontal scroll bar.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -35,6 +35,7 @@ h2 {
 
 .nook-phone {
   width: 100%;
+  max-width: 1400px;
   box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


mikebryant:
Looks good, thank you :)

Issue #43, opened 2020-04-08
Guaranteed Minimum?


Aliervo:
Just noticed that the final two columns are ``Guaranteed Minimum`` and ``Potential Maximum``.  I am assuming both are meant for the highest price of the week, correct?

If that is the case, perhaps we should rename the guaranteed min column to something like "Lowest Potential Maximum" and the current potential max to "Highest Potential Maximum".  Maybe it's just me, but "guaranteed minimum" seems to imply that turnips will not go below that price for the entire week.

rrsharpe:
It appears Guaranteed Minimum is the maximum of the lowest prices. This means you will definitely see at least that price some time during the week and should not sell at any price lower than that. It does seem a bit flawed however as it also includes past prices which are no longer accessible.

spfennell:
Yes, we need a column for future minimums, not ones that have already passed

PR #55, opened 2020-04-09
Add a guaranteed future minimum column

mgroat:
Implemented idea from https://github.com/mikebryant/ac-nh-turnip-prices/issues/43

The guaranteed minimum column could include a past value, this change will add another column that only figures in future values.

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -175,7 +175,7 @@ const calculateOutput = function (data, first_buy, previous_pattern) {
         out_line += `<td>${day.min}</td>`;
       }
     }
-    out_line += `<td>${poss.weekGuaranteedMinimum}</td><td>${poss.weekMax}</td></tr>`;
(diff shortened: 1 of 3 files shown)


js/scripts.js
```diff
@@ -175,7 +175,7 @@ const calculateOutput = function (data, first_buy, previous_pattern) {
         out_line += `<td>${day.min}</td>`;
       }
     }
-    out_line += `<td>${poss.weekGuaranteedMinimum}</td><td>${poss.weekMax}</td></tr>`;
+    out_line += `<td>${poss.weekGuaranteedMinimum}</td><td>${poss.futureMin}<td>${poss.weekMax}</td></tr>`;
```
Tekaoh, line 178: Seems to be missing a `</td>` here maybe

mikebryant:
Used #123 instead, but thank you for the contribution :)

PR #58, opened 2020-04-10
Added parameters first-time and previous-pattern.

Molorius:
Added text box to notify that values will not be saved.

Valid "first" options are:
yes/true/1
no/false/0.

Valid "pattern" options are:
0/fluctuating
1/large-spike
2/decreasing
3/small-spike
anything else defaults to unknown.

Test with:
https://turnipprophet.io/?pattern=large-spike&prices=90.91..93.94..&first=true

### DIFF
--- a/index.html
+++ b/index.html
@@ -273,6 +273,13 @@
     <p>This app is still in development, but will improve over time!</p>
   </div>
 
+  <div class="dialog-box" id="from_query" style="visibility: hidden;">
+    <h2 class="dialog-box__name">Daisy Mae</h2>
+    <p>I see you are have entered some parameters. 
+      I will not save any changes for now. 
+    </p>
+  </div>
+
   <div class="dialog-box">
     <h2 class="dialog-box__name">Daisy Mae</h2>
     <p>
--- a/index.html
+++ b/index.html
@@ -273,13 +273,6 @@
     <p>This app is still in development, but will improve over time!</p>
   </div>
 
-  <div class="dialog-box" id="from_query" style="visibility: hidden;">
-    <h2 class="dialog-box__name">Daisy Mae</h2>
-    <p>I see you are have entered some parameters. 
-      I will not save any changes for now. 
-    </p>
-  </div>
-
   <div class="dialog-box">
     <h2 class="dialog-box__name">Daisy Mae</h2>
     <p>
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -200,33 +200,26 @@ const getPricesFromQuery = function (param) {
 };
 
 const getPreviousFromQuery = function () {
-  const queries = [
-    ["first", getFirstBuyStateFromQuery],
-    ["pattern", getPreviousPatternStateFromQuery],
-    ["prices", getPricesFromQuery]
+
+  /* Check if valid prices are entered. Exit immediately if not. */
+  prices = getPricesFromQuery("prices");
+  if (prices == null) {
+    return null;
+  }
+
+  console.log("Using data from query.");
+  window.populated_from_query = true;
+  return [
+    getFirstBuyStateFromQuery("first"),
+    getPreviousPatternStateFromQuery("pattern"),
+    prices
   ];
-
-
-  found = null; /* value to save if we found any parameters */
-  ret = [];
-  for (q of queries) {
-    val = q[1](q[0]); /* run the function with the parameter */
-    found = found || val;
-    ret.push(val);
-  }
-
-  if (found != null) { /* if we found any parameter */
-    window.from_query = true;
-    document.getElementById("from_query").style.visibility = "visible";
-    return ret;
-  }
-  return null;
 };
 
 const getPreviousFromLocalstorage = function () {
   return [
     getFirstBuyStateFromLocalstorage(), 
-    getPreviousPatternStateFromQuery(),
+    getPreviousPatternStateFromLocalstorage(),
     getPricesFromLocalstorage()
   ];
 };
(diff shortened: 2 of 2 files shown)


Molorius:
#57 also adds the previous pattern parameter (with "previous_pattern"), and looks cleaner than my messy if/else statement.

tylermatchett:
Oh sorry, didn't check PRs before submitting my version of this.

Molorius:
Mine has the benefit that if any parameters are passed then it won't look to local storage for anything. Looks like several people were eager to implement this, so that's pretty cool!

tylermatchett:
Yeah ours both look similar in features, I'm just glancing through your code now

index.html
```diff
@@ -273,6 +273,13 @@ <h2 class="dialog-box__name">Daisy Mae</h2>
     <p>This app is still in development, but will improve over time!</p>
   </div>
 
+  <div class="dialog-box" id="from_query" style="visibility: hidden;">
+    <h2 class="dialog-box__name">Daisy Mae</h2>
+    <p>I see you are have entered some parameters. 
+      I will not save any changes for now. 
+    </p>
+  </div>
```
taylormarshall, line 281: Without taking a position on whether or not we should add this (I could see either way), it does seem to be a different feature than query support, likely makes sense as a different PR.

js/scripts.js
```diff
@@ -141,15 +193,52 @@ const getPricesFromQuery = function () {
...
+    document.getElementById("from_query").style.visibility = "visible";
+    return ret;
+  }
+  return null;
+};
+
+const getPreviousFromLocalstorage = function () {
+  return [
+    getFirstBuyStateFromLocalstorage(), 
+    getPreviousPatternStateFromQuery(),
```
taylormarshall, line 229: I think this line should pull from local storage, not from query

js/scripts.js
```diff
@@ -141,15 +193,52 @@ const getPricesFromQuery = function () {
...
+  ];
+};
+
+
+/**
+ * Gets previous values. First tries to parse parameters,
+ * if none of them match then it looks in local storage.
+ * @return {[first time, previous pattern, prices]}
+ */
+const getPrevious = function () {
```
taylormarshall, line 240: For me, packing up these 3 pieces of data makes it less readable.  I'd suggest leaving it as 3 separate functions (having a helper for each piece of data to either get from query or get from local storage seems nice though).

js/scripts.js
```diff
@@ -141,15 +193,52 @@ const getPricesFromQuery = function () {
...
+  ];
+
+
+  found = null; /* value to save if we found any parameters */
+  ret = [];
+  for (q of queries) {
+    val = q[1](q[0]); /* run the function with the parameter */
+    found = found || val;
+    ret.push(val);
+  }
```
taylormarshall, line 216: This is better off as just 3 function calls in a row, rather than building the list of queries and then unpacking it in this list.

js/scripts.js
```diff
@@ -187,13 +276,13 @@ const update = function () {
...
   const previous_pattern = parseInt(getCheckedRadio(previous_pattern_radios));
-
+  
   buy_input[0].disabled = first_buy;
   buy_input[0].placeholder = first_buy ? '—' : '...'
 
   const prices = [buy_price, buy_price, ...sell_prices];
 
-  if (!window.price_from_query) {
+  if (!window.from_query) {
```
taylormarshall, line 285: how about something a bit more descriptive,  'populated_from_query'  or similar?

index.html
Molorius: Fair enough, I'll remove it.

js/scripts.js
Molorius: That is a mistake, it should pull from local storage.

js/scripts.js
Molorius: I structured it that way because it didn't make sense to me to mix query data with local data. I made it this way so that if there are any query parameters, it won't look in local storage at all. Do you think it would be better to leave it messy, allow it to mix query data with local data (the old way), or look at a global flag?

js/scripts.js
Molorius: That was to make it dynamic in case we want to add more functions. I will split it.

js/scripts.js
taylormarshall: Agreed 100% that we shouldn't mix local storage with query string data.  I see what you were getting at with this structure now.

My first thought for how to do this would be to check for existing of query params and then decide which 3 functions to call.

On discord today, it was noted by mikebryant that we probably want prices to be required if query string is going to be used at all.  So you could just check for the existence of that param, and ignore the rest if it's not there.

Molorius: By leaving the helper function we will use local storage if the queries are invalid, such as "prices=test". Adding an extra check for that doubles up the work of the *FromQuery functions. I'm not sure how to get around that if we're calling them separately.

taylormarshall: I see what you mean.  I will leave it for folks who are approving PRs to decide, but from my POV if the url param is invalid, I would prefer to see some sort of error message, rather than silently loading the localstorage.   So I think it would be desirable to go down a completely different path if they a present at all.

Molorius: I won't add a UI piece here because that should probably be a separate pull request (as discussed above). I will add a console log for when we are using a query, that's probably appropriate for anyone designing one anyway.

mikebryant:
I like the encapsulation of the 3 pieces into a data structure. Thanks for the contribution!

PR #11, opened 2020-04-06
(WIP) Added data visualization

Charlignon:
Hello,

I really liked this tool, and I really like graphs too.
I thought about adding some.

This PR is just a draft so  you can tell me what you think of it !
If you are interested, I will enhance and clean the code.

Looking forward to your feedback !

PR #61, opened 2020-04-10
Add data visualization

PastisD:
I know that #11 add dataviz, but it's made on an old version :P

So I added the same graph as https://elxris.github.io/Turnip-Calculator/ on current version
The graph is based on "All patterns" probability.

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -189,6 +189,10 @@
 
       <h2>Output</h2>
 
+      <div class="form__flex-wrap">
+        <canvas id="chart" width="100%" height="400"></canvas>
+      </div>
+
       <div class="table-wrapper">
         <table id="turnipTable">
           <thead>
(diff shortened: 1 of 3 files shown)


rhuwellsamano:
Yes!!

theRTC204:
I like the idea of having some sort of data visualization, but lifting the exact code from another popular open source project, without providing reference back to the original creators, seems kind of scummy. 🤷‍♂️

PastisD:
Indeed, I forgot to add link in footer :)

mikebryant:
There's quite a few things I don't like about their version, so I'm afraid I won't be accepting this. Thank you for the contribution though - it's given me a poke, and I've started from scratch using Chart.js and made #104

PastisD:
Not a problem, i wanted a chart, you made it :P

PR #64, opened 2020-04-11
Provides consistent cross-browser sorting behavior.

phoenix-meadowlark:
Changes the sorting in `generate_possibilities` to have the intended effect. 

The old version worked on Firefox, but did not sort the list on Chrome because `a.weekMax < b.weekMax` does not return `-1` if `a.weekMax > b.weekMax`.


mikebryant:
Thanks for fixing this! :)

PR #77, opened 2020-04-13
Added favicon

vista178:
Would you like a favicon?


mikebryant:
Thanks for the contribution, but I'm afraid I ended up going with a different proposal for the icon.

PR #78, opened 2020-04-13
Clarify price should be from your island

timplunkett:
This is probably obvious for others, but at first I mistakenly put in the price I bought at, but it was from my friend's island. Just adding this text might help some people.


mikebryant:
Thanks for the contribution :)

Issue #72, opened 2020-04-12
[Feature Proposal] Overall Percent Chance


nbernabe09:
Hello,

I was wondering if this could be an additional feature to continue with the statistic trend.

![image](https://user-images.githubusercontent.com/28968121/79079268-ed013380-7cc2-11ea-8627-3c7576128169.png)

I made a `working sample` for the current version as of 4/12/2020. It loops through the data found in the table and adds all of the percent chances per category.

I use the `Custom JavaScript for Websites 2` extension for Chrome to run outside scripts into a page. If anyone wants to sample this in their own browser, you can use the following JavaScript code:

```js
var trend = {};

function getChances() {
  trend = {
    random: {
      chance: 0,
      count: 0,
      min: 0,
      max: 0
    },
    down: {
      chance: 0,
      count: 0,
      min: 0,
      max: 0
    },
    small: {
      chance: 0,
      count: 0,
      min: 0,
      max: 0
    },
    large: {
      chance: 0,
      count: 0,
      min: 0,
      max: 0
    }
  }
  $('#output tr').each(function(){
    var type = $(this).children('td:nth-child(1)').text();
    var percent = parseFloat($(this).children('td:nth-child(2)').text());
    var low = parseInt($(this).children('td:nth-child(16)').text());
    var high = parseInt($(this).children('td:nth-child(17)').text());
    
    switch (type) {
      case 'Fluctuating':
        type = trend.random;
        break;
      case 'Decreasing':
        type = trend.down;
        break;
      case 'Small spike':
        type = trend.small;
        break;
      case 'Large spike':
        type = trend.large;
        break;
    }
    
    type.chance = percent;
    type.count++;
    type.min = low;
    type.max = high;
  });
  
  var totalCount = trend.large.count + trend.small.count + trend.random.count + trend.down.count;
  var randomChance = isNaN(trend.random.chance) ? `${((trend.random.count / totalCount) * 100).toFixed(0)}%` : `${(trend.random.chance * trend.random.count).toFixed(0)}%`;
  var downChance = isNaN(trend.down.chance) ? `${((trend.down.count / totalCount) * 100).toFixed(0)}%` : `${(trend.down.chance * trend.down.count).toFixed(0)}%`;
  var smallChance = isNaN(trend.small.chance) ? `${((trend.small.count / totalCount) * 100).toFixed(0)}%` : `${(trend.small.chance * trend.small.count).toFixed(0)}%`;
[...]

PR #79, opened 2020-04-13
Sort row output by probability

### LINKED ISSUE #65: [Feature Proposal] Sort output rows by probability, and keep the "All patterns" row at the top.
Reading the output table could be more intuitive if the rows were sorted by their probability, so that the most likely outcome is the first one that the user sees.

It would also be nice keep the "All patterns" row at the top of the table as a summary.

If either of these changes sound good, I'd be happy to make a pull request referencing this issue that implements them.

Here's a side-by-side of the current and proposed behavior:
current_behavior

proposed_change

ChiefMilesEdgeworth:
[x] Sorts output by probability in descending order
[x] Sets the "All patterns" case at the top

Closes #65 

Let me know if there's anything that needs done! :)

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -654,15 +654,9 @@ function analyze_possibilities(sell_prices, first_buy, previous_pattern) {
     poss.weekMax = Math.max(...weekMaxes);
   }
 
-  generated_possibilities.sort((a, b) => {
-    if (a.weekMax < b.weekMax) {
-      return 1;
-    } else if (a.weekMax > b.weekMax) {
-      return -1;
-    } else {
-      return 0;
-    }
-  });
+  generated_possibilities.sort((a, b) => 
+    b.pattern_description == "All patterns" ? 1 : b.probability - a.probability
+  );
 
   return generated_possibilities;
 }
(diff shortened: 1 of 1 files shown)


theRTC204:
After thinking further on this, sorting by probability could be misleading. Given the probabilities are currently on a per-row basis, the more possible matches there are for one pattern, the lower it's individual row probability will be.

I think this change needs more thought before introduction.

ChiefMilesEdgeworth:
I wonder if it would be possible to have all the rows that share a probability showing just one large block probability

ChiefMilesEdgeworth:
I could see if I could do something like that

theRTC204:
There's discussion around overall % chance happening in issue https://github.com/mikebryant/ac-nh-turnip-prices/issues/72.

ChiefMilesEdgeworth:
I'll hold onto this until that discussion finalizes, and see if I can't sort it based on the large percentages similarly. :)

ChiefMilesEdgeworth:
If #115 is merged, this won't be necessary anymore.

ChiefMilesEdgeworth:
Actually, I'm just going to close this since it's probably not the behavior we want regardless.

PR #88, opened 2020-04-14
Deepen the color of the highest price

kum1ko:


js/scripts.js
```diff
@@ -168,9 +168,23 @@ const calculateOutput = function (data, first_buy, previous_pattern) {
...
+      return value
+    })
+
+
     for (let day of poss.prices.slice(1)) {
       if (day.min !== day.max) {
-        out_line += `<td>${day.min} to ${day.max}</td>`;
+        let background = "transparent"
+        if (day.max === totalMax) background = "#f443369c"
+        out_line += `<td style="background: ${background}">${day.min} - ${day.max}</td>`;
```
theRTC204, line 187: This should be achieved through a CSS class, instead of a style attribute.

js/scripts.js
```diff
@@ -168,9 +168,23 @@ const calculateOutput = function (data, first_buy, previous_pattern) {
   for (let poss of analyze_possibilities(data, first_buy, previous_pattern)) {
     var out_line = "<tr><td class='table-pattern'>" + poss.pattern_description + "</td>"
     out_line += `<td>${Number.isFinite(poss.probability) ? ((poss.probability * 100).toPrecision(3) + '%') : '—'}</td>`;
+
+    let datas    = poss.prices.slice(1)
+    let totalMax = 0
+    let totalMin = 0
+
+    datas = datas.map(value => {
```
theRTC204, line 176: `Array.map()` is not a great option here, as it creates and returns an entirely new Array instance. This should be a standard loop instead.

kum1ko: I just fixed it XD

js/scripts.js
kum1ko: I just fixed it XD

js/scripts.js
BekaD12: Hi kum1ko, you can use what already exist if i don't make any mistake.

if (day.min !== day.max) {
                out_line += `<td>${day.min} to <span class="${day.max === poss.weekMax ? 'highest-price' : ''}">${day.max}</span></td>`;
            }

mikebryant:
Thanks for the contribution! I've gone with a different PR I'm afraid, but I hope it gives you what you're looking for.

Keep enjoying the app, and may your turnips sell high!

PR #89, opened 2020-04-14
Add steady state possibility for unknown previous pattern

### LINKED ISSUE #68: Add probabilities when user does not know previous pattern
We can calculate the steady state probability of any given pattern using a [Markov model](https://en.wikipedia.org/wiki/Markov_model).  Thus, we can get _initial_ probabilities for the 4 states and display it to the users when they select "I don't know" for the "Previous Pattern".  This builds on issue #13.

If my calculations are correct (details below; someone please verify), given no information about the previous week and no information on Mon - Sat for the current week, the chances for each pattern are:
```
pattern0: 0.346
pattern1: 0.247
pattern2: 0.148
pattern3: 0.259
```

I'm using a [Markov chain](https://en.wikipedia.org/wiki/Markov_chain) with the pattern numbers & probabilities from [Ninji's `TurnipPrices::calculate()`](https://gist.github.com/Treeki/85be14d297c80c8b3c0a76375743325b#file-turnipprices-cpp-L135-L210).  And, since I've not done these calculations in over a decade, I'm using an [online calculator](https://kilin.clas.kitasato-u.ac.jp/software/markov/markov.html) with the following data:
```
0.20 0.30 0.15 0.35
0.50 0.05 0.20 0.25
0.25 0.45 0.05 0.25
0.45 0.25 0.15 0.15
```

I'm not certain how this is affected by issue #48, which seeks to update probabilities based on Bayes.  I suspect the problems are independent; that is, the solution for issue #48 will just work when given these probabilities.  If not, maybe a [HMM](https://en.wikipedia.org/wiki/Hidden_Markov_model) would work?

peter50216:
We can go further and ask for the previous known pattern and how many weeks have passed since then, but I don't think that would be useful.

Fix #68

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -3,6 +3,7 @@ const PATTERN = {
   LARGE_SPIKE: 1,
   DECREASING: 2,
   SMALL_SPIKE: 3,
+  UNKNOWN: 4,
 };
 
 const PATTERN_COUNTS = {
@@ -10,6 +11,7 @@ const PATTERN_COUNTS = {
   [PATTERN.LARGE_SPIKE]: 7,
   [PATTERN.DECREASING]: 1,
   [PATTERN.SMALL_SPIKE]: 8,
+  [PATTERN.UNKNOWN]: 0,
 }
 
 const PROBABILITY_MATRIX = {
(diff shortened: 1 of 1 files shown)


js/predictions.js
```diff
@@ -3,13 +3,15 @@ const PATTERN = {
...
   SMALL_SPIKE: 3,
+  UNKNOWN: 4,
 };
 
 const PATTERN_COUNTS = {
   [PATTERN.FLUCTUATING]: 56,
   [PATTERN.LARGE_SPIKE]: 7,
   [PATTERN.DECREASING]: 1,
   [PATTERN.SMALL_SPIKE]: 8,
+  [PATTERN.UNKNOWN]: 0,
```
ColMelvin, line 14: Does this not cause `row_probability()` to return `Infinity`?

peter50216:
Close in favor of #90

Issue #48, opened 2020-04-09
Probability may not be accurate


ZZZZzzzzac:
current algorithm calculates probability via redistribution probability based on user input price falling into some price intervals and rule out those impossible ones, right? But I think this way may not be accurate. (still a good approximation)

say 100 for buy price and 85 on MonAM and last week is pattern3, it rule out pattern0 and redistribute. Total probabilities are:
```
pattern1 = 0.25*1   /(0.25*1+0.15*7/8+0.15*1)=0.47/7=0.0672each
pattern2 = 0.15*7/8 /(0.25*1+0.15*7/8+0.15*1)=0.28
pattern3 = 0.15*1   /(0.25*1+0.15*7/8+0.15*1)=0.247/7=0.0353 each
```

But for pattern 1 and 2 (they follow same rule on Mon so treat as one), it needs : `rate = 0.85 for rate = randfloat(0.85,0.9)`
and for pattern3, it needs: `rate = 0.85 for rate = randfloat(0.4,0.9)`
clearly the probability for pattern3 to get rate = 0.85 is much smaller(10 times) than pattern 1 and 2

using bayes equation it should be:
```
pattern1 = 0.25*0.01/(0.9-0.85) /( 0.25*0.01/(0.9-0.85)+0.15*0.01/(0.9-0.4)+0.15*0.01/(0.9-0.85))=0.602/7=0.0860 each
pattern2 = 0.15*0.01/(0.9-0.85) /( 0.25*0.01/(0.9-0.85)+0.15*0.01/(0.9-0.4)+0.15*0.01/(0.9-0.85))=0.361
pattern3 = 0.15*0.01/(0.9-0.4)  /( 0.25*0.01/(0.9-0.85)+0.15*0.01/(0.9-0.4)+0.15*0.01/(0.9-0.85))=0.0361/7=0.0051 each
```

or generally 
`P(pattern is k | input price is X) = P( X | k) *P(k) / sum(P( X | k) *P(k)) k from 0 to 3`

key here is `P( X | k)`, under pattern_k find probability of input price X, i think assuming X is uniformly distributed in calculated maxmin interval is a good start, since most of turnip price is generated by `sellPrices[work++] = intceil(randfloat(A, B) * basePrice);`, only for decreasing pattern like 
```
rate = randfloat(0.9, 0.85);
    for (work = 2; work < peakStart; work++)
    {
      sellPrices[work] = intceil(rate * basePrice);
[...]

PR #90, opened 2020-04-14
feature: Add steady state pattern probabilities

ColMelvin:
Using linear algebra, calculate the steady state probability of the
PROBABILITY_MATRIX Markov chain to provide the chances of each pattern
occurring when the user does not know the previous weeks' pattern.

Calculations are done on the fly, at the cost of a little extra time, to
make probability distribution updates easier for future developers.  If
the chances are updated (because of an update or for a different game in
the series), no additional knowledge is required of the developer to
provide these steady states.

Closes: #68

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -12,32 +12,16 @@ const PATTERN_COUNTS = {
   [PATTERN.SMALL_SPIKE]: 8,
 }
 
-const PROBABILITY_MATRIX = {
-  [PATTERN.FLUCTUATING]: {
-    [PATTERN.FLUCTUATING]: 0.20,
-    [PATTERN.LARGE_SPIKE]: 0.30,
-    [PATTERN.DECREASING]: 0.15,
-    [PATTERN.SMALL_SPIKE]: 0.35,
-  },
-  [PATTERN.LARGE_SPIKE]: {
-    [PATTERN.FLUCTUATING]: 0.50,
-    [PATTERN.LARGE_SPIKE]: 0.05,
-    [PATTERN.DECREASING]: 0.20,
-    [PATTERN.SMALL_SPIKE]: 0.25,
-  },
-  [PATTERN.DECREASING]: {
-    [PATTERN.FLUCTUATING]: 0.25,
-    [PATTERN.LARGE_SPIKE]: 0.45,
-    [PATTERN.DECREASING]: 0.05,
-    [PATTERN.SMALL_SPIKE]: 0.25,
-  },
-  [PATTERN.SMALL_SPIKE]: {
-    [PATTERN.FLUCTUATING]: 0.45,
-    [PATTERN.LARGE_SPIKE]: 0.25,
-    [PATTERN.DECREASING]: 0.15,
-    [PATTERN.SMALL_SPIKE]: 0.15,
-  },
-};
+const PROBABILITY_MATRIX = [
+  /* from \ to FLUC  LRG^  DECR  SML^ */
+  /* FLUC */ [ 0.20, 0.30, 0.15, 0.35 ],
+  /* LRG^ */ [ 0.50, 0.05, 0.20, 0.25 ],
+  /* DECR */ [ 0.25, 0.45, 0.05, 0.25 ],
+  /* SML^ */ [ 0.45, 0.25, 0.15, 0.15 ],
+];
+
+// Calculated to ensure value is accurate if base probabilities ever change.
+const STEADY_STATE_PROBABILITY = find_steady_state(PROBABILITY_MATRIX);
 
 function minimum_rate_from_given_and_base(given_price, buy_price) {
   return 10000 * (given_price - 1) / buy_price;
@@ -596,14 +580,13 @@ function* generate_possibilities(sell_prices, first_buy) {
 }
 
 function row_probability(possibility, previous_pattern) {
+  if (typeof previous_pattern === 'undefined' || Number.isNaN(previous_pattern) || previous_pattern === null || previous_pattern < 0 || previous_pattern > 3) {
+    return STEADY_STATE_PROBABILITY[possibility.pattern_number] / PATTERN_COUNTS[possibility.pattern_number];
+  }
   return PROBABILITY_MATRIX[previous_pattern][possibility.pattern_number] / PATTERN_COUNTS[possibility.pattern_number];
 }
 
 function get_probabilities(possibilities, previous_pattern) {
-  if (typeof previous_pattern === 'undefined' || Number.isNaN(previous_pattern) || previous_pattern === null || previous_pattern < 0 || previous_pattern > 3) {
-    return possibilities
-  }
-
   var max_percent = possibilities.map(function (poss) {
     return row_probability(poss, previous_pattern);
   }).reduce(function (prev, current) {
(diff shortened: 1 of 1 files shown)


js/predictions.js
```diff
@@ -596,14 +577,15 @@ function* generate_possibilities(sell_prices, first_buy) {
 }
 
 function row_probability(possibility, previous_pattern) {
+  if (typeof previous_pattern === 'undefined' || Number.isNaN(previous_pattern) || previous_pattern === null || previous_pattern < 0 || previous_pattern > 3) {
+	// Calculated to ensure value is accurate if base probabilities ever change.
```
peter50216, line 581: nit: there's some inconsistent indent here, use space instead of tab?

js/predictions.js
```diff
@@ -596,14 +577,15 @@ function* generate_possibilities(sell_prices, first_buy) {
...
+	var STEADY_STATE_PROBABILITY = find_steady_state(PROBABILITY_MATRIX);
```
peter50216, line 582: Should this be done once and initialized as a global constant?

js/predictions.js
ColMelvin: Thanks, I thought I caught those; will fix.

js/predictions.js
ColMelvin: I agree; my experience with JavaScript scoping is limited and I ran into a few unexplained issues that I never resolved.  I can give it another go; suggestions welcome.

In my ideal world, the steady state would only be calculated once and it would only be calculated if the user selected `I don't know` for the previous pattern.

ColMelvin: @peter50216 What are you thoughts about the global variable?  Is that common in JavaScript?

peter50216: Just move this to under the `const PROBABILITY_MATRIX = ...` line, mark it as `const`, and move the calculation functions up should be good enough.

> In my ideal world, the steady state would only be calculated once and it would only be calculated if the user selected `I don't know` for the previous pattern.

I feel that we probably don't need the optimization of only calculate this on the `I don't know` case, since the calculation is not _that_ time-consuming, and having a static constant makes code easier to read then having some kind of lazy-calculated factory function.

peter50216: > @peter50216 What are you thoughts about the global variable? Is that common in JavaScript?

I'm perfectly fine with global **constant** variable (like this one).

RPigott:
Hey guys, I was just looking into adding this feature myself when I found it was being worked on.

Do you really think there is any value in including the routine for calculating the steady state from transition matrix? The constant steady-state vector is trivial to calculate with common utilities like numpy, matlab, R, julia... It would be comparatively big reduction in code complexity here to just include the result of that calculation and not the procedure. Easier to review and maintain.

One component I thought might be a part of this feature is allowing the user to just input the previous week prices themselves if they don't know their pattern. In the case where the user does not know enough of the previous week prices to confidently identify the pattern, (like for example only last saturday's prices) we still have enough information to eliminate some patterns and make a better informed estimate of the chance of the current pattern by using the transition matrix in the obvious way: left multiply the calculated match probabilities.

E.g. if the user knows previous saturday prices that indicate a 40% chance of a high-spike pattern and a 60% chance of a fluctuating pattern, the product of the transition matrix and the known probabilities, [0.6, 0.4, 0, 0], gives the best known estimate of the pattern chances for the current week.

peter50216:
> Do you really think there is any value in including the routine for calculating the steady state from transition matrix? The constant steady-state vector is trivial to calculate with common utilities like numpy, matlab, R, julia... It would be comparatively big reduction in code complexity here to just include the result of that calculation and not the procedure. Easier to review and maintain.
> 

My feeling is that in case there's change in probability matrix (due to update, ...), it's easier to just update that and let the code to calculate the new steady state instead of having instruction to ask people manually updating the steady state probability.

Also the code to calculate the steady state is not that much and quite clean / straight-forward in my opinion, so I'm leaning toward this rather than hard-coded the probability (#89).

> One component I thought might be a part of this feature is allowing the user to just input the previous week prices themselves if they don't know their pattern. In the case where the user does not know enough of the previous week prices to confidently identify the pattern, (like for example only last saturday's prices) we still have enough information to eliminate some patterns and make a better informed estimate of the chance of the current pattern by using the transition matrix in the obvious way: left multiply the calculated match probabilities.
> 
> E.g. if the user knows previous saturday prices that indicate a 40% chance of a high-spike pattern and a 60% chance of a fluctuating pattern, the product of the transition matrix and the known probabilities, [0.6, 0.4, 0, 0], gives the best known estimate of the pattern chances for the current week.

Yeah this feature looks really useful, but I feel that it can be a separate PR. Open a feature request for this?
Would also need #48 to be solved before we can actually get accurate probability though.

ColMelvin:
> Do you really think there is any value in including the routine for calculating the steady state from transition matrix? The constant steady-state vector is trivial to calculate with common utilities like numpy, matlab, R, julia... It would be comparatively big reduction in code complexity here to just include the result of that calculation and not the procedure. Easier to review and maintain.

Alternatively, requiring authors to calculate steady states might limit who can update the probabilities.  Moreover, conventional wisdom with RDBMS systems is to never store derived values, because data gets out of sync and can cause maintenance headaches later.

That said, I definitely agree: less code generally correlates with fewer bugs.  And this is a rather inconsequential piece of code; inaccuracy is probably acceptable if the steady state is not updated.

I leave it to the community/maintainer to decide if #89 or #90 makes more sense based on their views of "maintenance burden".  I had fun re-learning the Gauss-Jordan elimination, so I'm happy.

> One component I thought might be a part of this feature is allowing the user to just input the previous week prices themselves if they don't know their pattern. In the case where the user does not know enough of the previous week prices to confidently identify the pattern, (like for example only last saturday's prices) we still have enough information to eliminate some patterns and make a better informed estimate of the chance of the current pattern by using the transition matrix in the obvious way: left multiply the calculated match probabilities.
> 
> E.g. if the user knows previous saturday prices that indicate a 40% chance of a high-spike pattern and a 60% chance of a fluctuating pattern, the product of the transition matrix and the known probabilities, [0.6, 0.4, 0, 0], gives the best known estimate of the pattern chances for the current week.

I thought about the idea as well.  However, if there is more than one possible previous pattern, we need much more involved modelling.  Indeed, the first place I would look is to a [HMM](https://en.wikipedia.org/wiki/Hidden_Markov_model).

js/predictions.js
ColMelvin: Changed.  I guess something else - elsewhere - was wrong when I tried earlier.

RPigott:
> Alternatively, requiring authors to calculate steady states might limit who can update the probabilities. Moreover, conventional wisdom with RDBMS systems is to avoid storing derived values, because data can get out of sync and can cause maintenance headaches later.

I'm not going to push too hard for one approach or another here, you guys do whatever you want with it. But as a final justification for my opinion, I believe it will not be helpful to prepare for a theoretical update that changes only the transition probabilities and which would already require manual intervention to correct for. As for limiting who could update it, I'll offer the following julia one-liner to calculate the desired chances:

```julia
using LinearAlgebra
p = [ 0.20 0.50 0.25 0.45 
      0.30 0.05 0.45 0.25
      0.15 0.20 0.05 0.15
      0.35 0.25 0.25 0.15 ]
steady = normalize(nullspace(I - p) |> vec, 1)
```
You could even treat the steady-state as the transition probability from an "unknown" pseudo-state in the transition matrix and store the values there if you are worried they would be forgotten. The transitions
```julia
[0; [steady p]]
5×5 Array{Float64,2}:
 0.0       0.0   0.0   0.0   0.0
 0.346277  0.2   0.5   0.25  0.45
 0.247363  0.3   0.05  0.45  0.25
 0.147607  0.15  0.2   0.05  0.15
 0.258752  0.35  0.25  0.25  0.15
```
now describe the evolution from ["unknown", "fluctuating", "high-spike", "decreasing", "small-spike"] to the current pattern. Obviously, it is impossible to transition into an unknown state, and an unknown state becomes a "known" uncertain state. This eliminates the need for any special-casing of an unknown pattern.

> > In the case where the user does not know enough of the previous week prices to confidently identify the pattern, [we still have enough information to make an informed estimate about the current pattern] **by using the transition matrix in the obvious way: left multiply the calculated match probabilities.**

> I thought about the idea as well. However, if there is more than one possible previous pattern, we need much more involved modelling. Indeed, the first place I would look is to a HMM.

No, the model remains exactly the same. There are no hidden states. The transition matrix completely characterizes the evolution of a mixed/uncertain state to a new uncertain state. Since we have no information of the pattern 2 weeks before, the prior probabilities (before the bayesian correction from the user's info) of the previous week's patterns should be taken to be the steady-state values. Left multiply last week's posterior probabilities by the transition matrix to yield the current week's pattern chances.

It's true though that #48 will need to be addressed first. On quick review, I think there are additional inaccuracies other than those described in #48 that we could also improve on. Per suggestion, I'll open a new request/pr for this feature and related issues at a later date.

ColMelvin:
Closing as unnecessary; issue #68 fix merged in #109.

PR #91, opened 2020-04-14
fix: Removed a debug statement left behind

theRTC204:
Accidentally left a debug statement in the contributors script. 😬


mikebryant:
Thanks :)

PR #93, opened 2020-04-14
feat: Implements a more true to game code intceil calculation

theRTC204:
This replaces all instances of `Math.floor()` and `Math.ceil()` with a calculation that more closely matches the C++ implementation in game. In the extracted code, the all values are passed through a function `intceil()` which is implemented as...

```cpp
int intceil(float val)
{
    return (int)(val + 0.99999f);
}
```
In C++ the behaviour of casting to an int results in a truncated value, not a rounded value. As such, there are situations where using JavaScript's `Math.floor()` or `Math.ceil()` can result in a value that is off by one. By using `Math.trunc()`, we can more closely represent the C++ behaviour.

I know we already have reports of calculation issues, and there is a feeling that having a wider net of possible values can reduce the number of failed predictions, but I personally feel the wider net makes it more difficult to diagnose potential problems in this app's calculations. I feel it would be best to represent the in-game calculation as closely as possible.

Open to discussion of course!

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -39,6 +39,10 @@ const PROBABILITY_MATRIX = {
   },
 };
 
+function intceil(val) {
+  return Math.trunc(val + 0.99999);
+}
+
 function minimum_rate_from_given_and_base(given_price, buy_price) {
   return 10000 * (given_price - 1) / buy_price;
 }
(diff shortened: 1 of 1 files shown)


mikebryant:
Sure, works for me

PR #102, opened 2020-04-15
Update rate range to match intceil

peter50216:
I doubt this would change any prediction at all, but this is more correct according to the code.


mikebryant:
Sounds good, thanks :)

PR #103, opened 2020-04-15
Change prediction of rate to consider not only the previous known value

peter50216:
This also makes the predicted range more strict, so won't help with all those invalid pattern issues, but should give an more accurate result.


peter50216:
Also checked this on the ~40 data rows (some of them are partial data) I have on hand and see no issue.

mikebryant:
This is a good idea, thanks :)

Issue #95, opened 2020-04-14
Inaccurate Probabilities: Update Probabilities based on Multiplier


spfennell:
This may be a duplicate of #48, but I'm making a new issue because I have long example and I don't want to crowd that issue

Probabilities are not always accurate in Turnip Prophet. I’ll give an example.

Scenario: 

Last week was Decreasing. So according to Ninji:
**P(Fluctuating) = 0.25
P(Large Spike) = 0.45
P(Decreasing) = 0.05
P(Small Spike) = 0.25**

Okay, so lets move forward to Monday morning:
Sunday = 94
Monday AM = 83

So our Monday rate is:
R = (Monday AM)/(Sunday) = 83/94 = 0.88

R = 0.88

Lets say if R >= 0.85 and R <= 0.90, we call this event A

Event A is a pretty common occurrence on the stalk market. It Always happens at the start of Large Spike and Decreasing patterns. Updating our probabilities based on event A is important to have accurate chance calculations.

**P(A | Fluctuating)  = 0
P(A | Large Spike) = 1
P(A | Decreasing)  = 1**

P(A | Small Spike) = ?
---------

Starts to get a little complicated here. 

Small spike starts with [0 to 7] days of decreasing prices starting at R = [0.4 to 0.9]

So the probability that we start with decreasing is 7/8, because we could get 0 days.

Suppose this is true, then the probability that R will be 0.85 ≤ R ≤ 0.9 is 
(0.9 - 0.85) / (0.9 - 0.4) = 0.1

So P(A | Small Spike) = 0.1 * 7 / 8 = 0.0875

**P(A | Small Spike) = 0.0875**

Now we will find the probability of A.
---------------
P(A) = P(A | Fluctuating)*P(Fluctuating) + P(A | Large Spike)*P(Large Spike) + P(A | Decreasing)*P(Decreasing) + P(A | Small Spike)*P(Small Spike)

= 0 + 0.45 + 0.05 + 0.0875*0.25 = 0.521875

**P(A) = 0.521875**

So let’s recap:

**P(Fluctuating) = 0.25
P(Large Spike) = 0.45
P(Decreasing) = 0.05
P(Small Spike) = 0.25
P(A | Fluctuating)  = 0
P(A | Large Spike) = 1
P(A | Decreasing)  = 1
P(A | Small Spike) = 0.0875
P(A) = 0.521875**

To get our new probabilities for each pattern, we need
P(Fluctuating | A) = ?
[...]

PR #105, opened 2020-04-15
Accurate probability.

### LINKED ISSUE #48: Probability may not be accurate

peter50216:
Tested that the result is same as the "should report" in #95.

Fix #48, #95.

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -381,13 +381,13 @@ function generate_peak_price(
   //
   // => X->U(A,B), Y->U(C,X), Y-C->U(0,X-A), Y-C->U(0,1)*(X-A), Y-C->U(0,1)*U(C-A,B-A),
   // let Z=Y-C,  Z1=C-A, Z2=B-A, Z->U(0,1)*U(Z1,Z2)
-  // Prob(Z>=t) = integral_{x=0}^{1} [min(t/x,Z2)-min(t/x,Z1)]/ (Z2-Z1)
+  // Prob(Z<=t) = integral_{x=0}^{1} [min(t/x,Z2)-min(t/x,Z1)]/ (Z2-Z1)
   // let F(t, ZZ) = integral_{x=0}^{1} min(t/x, ZZ)
   //    1. if ZZ < t, then min(t/x, ZZ) = ZZ -> F(t, ZZ) = ZZ
   //    2. if ZZ >= t, then F(t, ZZ) = integral_{x=0}^{t/ZZ} ZZ + integral_{x=t/ZZ}^{1} t/x
   //                                 = t - t/ZZ log(t/ZZ)
-  // Prob(Z>=t) = (F(t, Z2) - F(t, Z1)) / (Z2 - Z1)
-  // Prob(Y>=t) = Prob(Z>=t-C)
+  // Prob(Z<=t) = (F(t, Z2) - F(t, Z1)) / (Z2 - Z1)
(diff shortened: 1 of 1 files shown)


mikebryant:
I am going to blindly assume the probability stuff is right :D

Predictions still look good. Thanks for this, this is awesome :)

ZZZZzzzzac:
I should pin out that in that example in #95, the event A is still price X fall into interval (85,90) assuming base price is 100, so PR #105 actually not  `Make probability accurate by condition probability on the "value"` but on the range

peter50216:
> I should pin out that in that example in #95, the event A is still price X fall into interval (85,90) assuming base price is 100, so PR #105 actually not `Make probability accurate by condition probability on the "value"` but on the range

I indended to write "condition probability on the known value in range" but that's too long for commit message :P

Also the new PR is at https://github.com/mikebryant/ac-nh-turnip-prices/pull/109 (since this one has some bug and is reverted), please help take a look to see if my calculations are correct, thanks!

PR #116, opened 2020-04-15
Added code to generate permalink

joseamirandavelez:


js/permalink.js
```diff
@@ -0,0 +1,9 @@
+function getPermalink() {
+	if (window.jQuery) {
+		const container = $('#permalink');
+		if (permalink != null) {
+			container.append(permalink);
+		}
+	}
+}
+$(document).ready(getPermalink);
```
mikebryant, line 9: It's not clear to me what this whole file/function is needed for - on document ready we already call `initialize` which calls `update` - which will then set the permalink through the new code in `update`

js/scripts.js
```diff
@@ -272,6 +272,16 @@ const update = function () {
...
   const previous_pattern = parseInt(getCheckedRadio(previous_pattern_radios));
+  const sp = sell_prices.toString().replace(/NaN/g,"0").replace(/,/g,".");
+  const url = location.protocol + "//" + location.host + location.pathname + "?" +  sp + "&first=" + first_buy + "&pattern=" + previous_pattern;
+  const permalink_text = "Permalink";
+  const permalink = "<a href='" + url + "'>"+ permalink_text + "</a>" ;
+  //Permalink structure: /?prices=97.86.82.78.75.71.67.108.0.0.0.0.0&first=true&pattern=3
+
+  if (permalink != null) {
+	const container = $('#permalink');
+    container.append(permalink);
```
mikebryant, line 283: As you enter values, this results in a dialog box at the bottom with PermalinkPermalinkPermalinkPermalink - it should probably be resetting the element, not appending each time

js/permalink.js
joseamirandavelez: Its not needed. I removed it.

js/scripts.js
joseamirandavelez: Yes. I didnt actually know everything is updated real time. I fixed this by changing `append` to `html`.

joseamirandavelez:
This may probably need to be moved to a better location in index.html... Or maybe add a share button...

js/permalink.js
mikebryant: Thanks :)

mikebryant:
> This may probably need to be moved to a better location in index.html... Or maybe add a share button...

Yeah, this is workable now, but not particularly visually appealing. I'm not great at that side of things.. I'll poke someone else in the discord, see if anyone wants to take a stab at styling it better

theRTC204:
I can take a stab at styling this up a bit differently. I think the link should be up in the form probably above the graph - maybe with a copy to clipboard button next to it.

mikebryant:
@joseamirandavelez Thanks for kicking this off!

Permalink button is now live :)

joseamirandavelez:
> @joseamirandavelez Thanks for kicking this off!
> 
> Permalink button is now live :)

It seems youn modified the code, the permalink looks like this now:

https://turnipprophet.io/?prices=93...115.127.95.......&pattern=3

The zeros are not written. Still works, though.

joseamirandavelez:
Did you add the code on a different fork? I dont see my name in the contributors list.

PR #122, opened 2020-04-15
Add ACNH UI inspired hover effects for buttons and inputs

cmargroff:


cmargroff:
- I did light mobile testing. Mobile hover effects obviously aren't available. I originally tested having the scrolling background on selected but it was too much on screen at once. I don't think its feasible to implement this animation on mobile.

- I simply modified their :hover styles, they have a there are :placeholder-shown clauses on the input rules. I didn't want to change original functionality just styles.

PR #123, opened 2020-04-15
Future Minimum Price

Tekaoh:
Closes mikebryant/ac-nh-turnip-prices#43

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -570,8 +570,15 @@ function analyze_possibilities(sell_prices, first_buy, previous_pattern) {
     var weekMins = [];
     var weekMaxes = [];
     for (let day of poss.prices.slice(2)) {
-      weekMins.push(day.min);
-      weekMaxes.push(day.max);
+      // Check for a future date by checking for a range of prices
+      if(day.min !== day.max){
+        weekMins.push(day.min);
+        weekMaxes.push(day.max);
+      } else {
+        // If we find a set price after one or more ranged prices, the user has missed a day. Discard that data and start again.
+        weekMins = [];
+        weekMaxes = [];
+      }
     }
     poss.weekGuaranteedMinimum = Math.max(...weekMins);
     poss.weekMax = Math.max(...weekMaxes);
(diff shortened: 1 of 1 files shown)


Tekaoh:
Ah, this seems to be sort of a duplicate of #55. Sorry about that

theRTC204:
Sort of, yes, but the contributor of that one has been inactive for about a week and their PR has conflicts and cannot be merged. So, your work is still viable!

js/predictions.js
```diff
@@ -569,12 +569,18 @@ function analyze_possibilities(sell_prices, first_buy, previous_pattern) {
   for (let poss of generated_possibilities) {
     var weekMins = [];
     var weekMaxes = [];
+    var futureMins = [];
     for (let day of poss.prices.slice(2)) {
       weekMins.push(day.min);
       weekMaxes.push(day.max);
+      // Check if the date is in the future by checking for a range of prices
+      if(day.min !== day.max){
```
mikebryant, line 577: If someone misses a day, that will be counted for the future minimum - but it shouldn't be

index.html
```diff
@@ -246,7 +246,8 @@ <h2>Output</h2>
                   <span>PM</span>
                 </div>
               </th>
-              <th valign="bottom">Guaranteed Minimum</th>
+              <th valign="bottom">Weekly Minimum</th>
```
mikebryant, line 249: Is it worth keeping the weekly minimum column? Past prices aren't that useful.

js/predictions.js
mikebryant: Also, should we include today's value, or not - that's still something they're able to sell for

index.html
theRTC204: My 2 cents: Personally do not see myself using the other minimum column at all after this goes out.

js/predictions.js
theRTC204: I'm back and forth on including the current period value... On the one hand as you say it's still a valid sellable price, but on the other hand someone might misunderstand that as a possible value for tomorrow and then think waiting for tomorrow is worth the risk only to find out their min is actually lower.

Tekaoh: In my opinion, no. It might give the impression at a glance that today's price might come up again later, but it might not. I think it's useful as a comparison against today's price, which presumably the user already has in mind. But that's just me

index.html
Tekaoh: I agree. Just the one is probably better

Tekaoh:
Another question... Should the maximum column just show future prices too? Then the two columns would be consistent. Also, a user might see the the potential maximum price and try holding out for it, not realizing that perhaps that price was in the past.

Tekaoh:
Yes, no problem. That's a lot better @theRTC204

Tekaoh:
If this merges, it also Closes #55 

(Not sure if the auto link works from PR to PR, but I wanted to find out, hah)

PR #125, opened 2020-04-15
Color coding for price ranges

lluk:
Simple color coding for price ranges in the grid. Based on the maximum potential of a range, makes it easy to visualize which days you want to be interested in.

colored-ranges


theRTC204:
I like it! I wonder if blue & green are maybe reversed though? It seems a bit weird that blue is the best value, when green typical indicates "Go".

lluk:
There's actually a couple shades of red (for worse and worst) and a couple shades of green. I found it was harder to colour the large spikes with green still and have it still stand out. The blue seemed like a good way to highlight the best possible result since it appears rarely.

lluk:
Adding a live version here incase anyone wants to see it, finding the colouring more use than the graph.
https://lluk.github.io/ac-nh-turnip-prices/

akccakcctw:
Maybe we can calc color from diff (`max - buy_price`) by using `hsl()`?
```javascript
// red (0°), green (120°), blue (240°)
  const getHslByPriceDiff = function(diff) {

  let hue;
  if(diff < -30) hue = 0; // red
  if(diff >= -30) hue = 30;
  if(diff >= 0) hue = 60;
  if(diff >= 30) hue = 90;
  if(diff >= 200) hue = 140; // green - blue (more highlight than green)

  return `hsl(${hue}, 80%, 85%)`;
}
```

if too many `if` is ugly:
```javascript
const getHslByPriceDiff = function(diff) {
  // red (0°), green (120°), blue (240°)
  const diff2HueMapping = {
    '-30': 30,
    0: 60,
    30: 90,
    200: 140,
  };
  // just want to get the keys of the mapping...
  const hueList = Object.keys(diff2HueMapping).map(i => +i).sort((a, b) => a - b);

  if(diff < hueList[0]) return `hsl(0, 80%, 85%)`; // red

  const hue = diff2HueMapping[hueList.reduce((acc, cur) => (diff >= cur) ? cur : acc, diff)];
  return `hsl(${hue}, 80%, 85%)`;
}

const getPriceColor = function(buy_price, max) {
  return getHslByPriceDiff(max - buy_price);
}
```

Snipaste_2020-04-23_02-04-50

theRTC204:
My eyes tend to enjoy the pastel colors from @akccakcctw more.

mikebryant:
Mhmm, now that is pretty, I like :)

Do we need to do something special for dark mode?

lluk:
I made the following changes based on the suggestions from @akccakcctw :
- price class is now calculated based on `(max_price - buy_price)`
- used `hsl()` for the colour palette but it's since they're static hues, it's in the CSS instead of calculated
- the colours got inverted in dark mode (i.e. green became red) - so the CSS includes dark mode overrides to invert them back

live @ https://lluk.github.io/ac-nh-turnip-prices/

lluk:
Sorry, I squashed some commits.

Issue #70, opened 2020-04-12
Dark Mode


Raemistar:
The amount of white space is near blinding when playing the Stalk Market late at night. I would love to have a feature to enable dark mode!

makusujp:
maybe work with [https://darkmodejs.learn.uno/](https://darkmodejs.learn.uno/) ?

PR #127, opened 2020-04-16
darkmode using darkmodejs

Sujeom:
In reference to issue #70 

Lmk if color changes are needed or feel free to make any edits. :)


Sujeom:
Made some adjustments. this is the current look. Should I go for another scheme instead of the purple/navy?
Screenshot_20200416_011112

theRTC204:
I did see the adjustments... but I have to be honest I'm not really fond of the pallette used here. Personally I would like to see the site get a really eye catching darkmode overall.

I think I would like to get @mikebryant's opinion on this before I give it my approval.

Sujeom:
I get what you mean. I'm not entirely sure what pallette would give the best AC vibe.

mikebryant:
The blue/black theme is reasonable. I'm not sure what an "AC" theme would be. The note text (like "This affects your pattern") is a little hard to read for me.

Sujeom:
Updated the coloring on the note texts.

PR #109, opened 2020-04-15
Accurate probability (act 2)

peter50216:

PR #132, opened 2020-04-16
Fixed predictions in the Small Spike pattern

benru:
There were some errors in the min/max values for the peak in the small spike patterns. This change ensures that the peak is always greater than its sides, and the sides are always at least one less than the peak. The previous behaviour was almost correct as long as you had all values entered.

Examples of incorrect behaviour fixed by this PR:
- [105.74.112.100.........](https://turnipprophet.io?prices=105.74.112.100.........)
   The minimum for the peak (Wed AM) should be 147, not 146 as it is always higher than the sides.
- [105.74.112.100..160.......](https://turnipprophet.io?prices=105.74.112.100..160.......)
   The left side of the peak should also be clamped like the right side to ensure it is below the peak.
- [105.74.112.100...160......](https://turnipprophet.io/?prices=105.74.112.100...160......)
   The minimum for the peak should be 161 because it must be greater than its right side.
- [105.74.112.100.165.160.......](https://turnipprophet.io/?prices=105.74.112.100.165.160.......)
   Patterns where the sides are greater (or equal to) the peak should be rejected as invalid.

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -446,117 +446,88 @@ class Predictor {
 
     const buy_price = given_prices[0];
     let prob = 1;
-    let rate_range = [rate_min, rate_max];
+    let peak_range = [rate_min, rate_max];
+    let side_range = [rate_min, rate_max];
+    let min_pred = this.get_price(rate_min, buy_price);
+    let max_pred = this.get_price(rate_max, buy_price);
 
-    // * Calculate the probability first.
-    // Prob(middle_price)
-    const middle_price = given_prices[start + 1];
-    if (!isNaN(middle_price)) {
-      const min_pred = this.get_price(rate_min, buy_price);
-      const max_pred = this.get_price(rate_max, buy_price);
-      if (middle_price < min_pred - this.fudge_factor || middle_price > max_pred + this.fudge_factor) {
+    // First fill the ranges normally
+    for (let i = 0; i < 3; i++) {
+      if (i == 1) {
+        predicted_prices.push({
+          min: min_pred,
+          max: max_pred,
+        });
+      }
+      else {
+        // Side prices are always one less than the peak
+        predicted_prices.push({
+          min: min_pred - 1,
+          max: max_pred - 1,
+        });
+      }
+    }
+
+    // If we have a value for the peak, apply it now and adjust side maximums
+    let given_peak_price = given_prices[start + 1];
+    if (!isNaN(given_peak_price)) {
+      min_pred = predicted_prices[start + 1].min;
+      max_pred = predicted_prices[start + 1].max;
+
+      if (given_peak_price < min_pred - this.fudge_factor || given_peak_price > max_pred + this.fudge_factor) {
         // Given price is out of predicted range, so this is the wrong pattern
         return 0;
       }
+
       // TODO: How to deal with probability when there's fudge factor?
       // Clamp the value to be in range now so the probability won't be totally biased to fudged values.
-      const real_rate_range =
-          this.rate_range_from_given_and_base(clamp(middle_price, min_pred, max_pred), buy_price);
-      prob *= range_intersect_length(rate_range, real_rate_range) /
-        range_length(rate_range);
+      const real_peak_range =
+          this.rate_range_from_given_and_base(clamp(given_peak_price, min_pred, max_pred), buy_price);
+      prob *= range_intersect_length(peak_range, real_peak_range) /
+        range_length(peak_range);
       if (prob == 0) {
         return 0;
       }
 
-      rate_range = range_intersect(rate_range, real_rate_range);
+      predicted_prices[start + 1].min = given_peak_price;
+      predicted_prices[start + 1].max = given_peak_price;
+
+      side_range[1] = real_peak_range[1];
+      predicted_prices[start].max = given_peak_price - 1;
+      predicted_prices[start + 2].max = given_peak_price - 1;
     }
 
-    const left_price = given_prices[start];
-    const right_price = given_prices[start + 2];
-    // Prob(left_price | middle_price), Prob(right_price | middle_price)
-    //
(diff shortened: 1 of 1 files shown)


benru:
I have updated this PR to take into account the probability changes from #109

benru:
@mikebryant Are you able to take a look at this?

Here is an example that caught us out this week:
https://turnipprophet.io/?prices=101.70.65.60.56.52.48....151..&pattern=0
https://benru.github.io/ac-nh-turnip-prices/?prices=101.70.65.60.56.52.48....151..&pattern=0

The current version implies that prices could have been higher on Fri AM, which is not possible in either of the two small spike patterns. It also gets the probabilities wrong between the two patterns.

mikebryant:
I'll take a look.

This isn't something that concerns me greatly - the aim of this tool is to predict the future, prices before ones that were entered aren't relevant, as you can't time travel backwards to see them. (As far as I'm aware, that causes the prices to be regenerated afresh.)

Can understand this causes confusion though, will try and get this merged :)

mikebryant:
@peter50216 Can you give me a hand reviewing this?

peter50216:
This would need to be rebased on #277 

> Here is an example that caught us out this week:
> https://turnipprophet.io/?prices=101.70.65.60.56.52.48....151..&pattern=0
> https://benru.github.io/ac-nh-turnip-prices/?prices=101.70.65.60.56.52.48....151..&pattern=0
> 
> The current version implies that prices could have been higher on Fri AM, which is not possible in either of the two small spike patterns. It also gets the probabilities wrong between the two patterns.

I feel the current conditional probability is more likely to be correct, since 50/50 to both patterns while a value is given conditioning on different range doesn't seems likely to me.

Also we don't do backward prediction in all other places too (For example, https://turnipprophet.io?prices=100.80.77..72........&pattern=3, Note that the Tuesday AM range is lower than PM price), and I really don't think it's worthwhile to fix them since past prices are useless.

The off-by-one on Saturday morning price (which is a future price) is definitely worth a fix though.

Also we should think of a way to avoid user confusion here, probably by adding some warnings on this when there's missing price, or having the past prices gray out?

benru:
> I feel the current conditional probability is more likely to be correct, since 50/50 to both patterns while a value is given conditioning on different range doesn't seems likely to me.

The ranges aren't different for the two possible small spikes: either the entered value is the spike itself and the probability is 1/(202-142+1) or it is the slope the day before the spike and the probability is 1/(201-141+1). Both equal 1/61 so a 50%/50% split is correct.

> Also we don't do backward prediction in all other places too (For example, https://turnipprophet.io?prices=100.80.77..72........&pattern=3, Note that the Tuesday AM range is lower than PM price), and I really don't think it's worthwhile to fix them since past prices are useless.

My specific short spike example may not confuse expert users who understand the game's patterns and fill in values for every day of the week, but two weeks in a row I've had to explain to someone who either missed a day or their shop was closed that they didn't miss the peak even though the chart implies there's a chance they did.

Incidentally even though past prices are useless, by not doing backwards prediction you are missing out on some accuracy in your probabilities. Take a similar version to your example [https://turnipprophet.io?prices=100.80..70.........](https://turnipprophet.io?prices=100.80..70.........) which predicts Short Spike at 66% chance. The code currently gives the ranges for the missing day as 69-76 if Fluctuating, and 74-78 if Small Spike. However if you try and put in those numbers manually for the missing day you'll see that the valid ranges are actually 73-76 for Fluctuating, and that Small Spike is only possible for 75. This means that Small Spike is overrepresented by 5x and Fluctuating by 2x. The correct probabilities are 56% Fluctuating and only 44% Short Spike. That is a huge difference and I may investigate it in a separate PR.

peter50216:
> The ranges aren't different for the two possible small spikes: either the entered value is the spike itself and the probability is 1/(202-142+1) or it is the slope the day before the spike and the probability is 1/(201-141+1). Both equal 1/61 so a 50%/50% split is correct.

The price for the day before the spike is NOT uniform in that range, so this is not correct. See #170, and code comments on #277.

> Incidentally even though past prices are useless, by not doing backwards prediction you are missing out on some accuracy in your probabilities. Take a similar version to your example https://turnipprophet.io?prices=100.80..70......... which predicts Short Spike at 66% chance. The code currently gives the ranges for the missing day as 69-76 if Fluctuating, and 74-78 if Small Spike. However if you try and put in those numbers manually for the missing day you'll see that the valid ranges are actually 73-76 for Fluctuating, and that Small Spike is only possible for 75. This means that Small Spike is overrepresented by 5x and Fluctuating by 2x. The correct probabilities are 56% Fluctuating and only 44% Short Spike. That is a huge difference and I may investigate it in a separate PR.

I'm not 100% sure about this case, but the current code calculates the probability `P(a1,a2,a3,...)` (where ai are the prices) by `P(a1)*P(a2|a1)*P(a3|a1,a2)*....`, so it *should* be correct even when it only looks forward. But yeah this case is worth investigating, I feel the easiest way is to run a local simulation to generate the patterns, which would catch the error if the current probability is off by a lot. I'll do this later.

peter50216:
> I'm not 100% sure about this case, but the current code calculates the probability `P(a1,a2,a3,...)` (where ai are the prices) by `P(a1)*P(a2|a1)*P(a3|a1,a2)*....`, so it _should_ be correct even when it only looks forward. But yeah this case is worth investigating, I feel the easiest way is to run a local simulation to generate the patterns, which would catch the error if the current probability is off by a lot. I'll do this later.

Did a quick simulation on this one:
Simulation data same as https://turnipprophet.io?prices=100.80..70.........&pattern=0
Simulation code: https://gist.github.com/peter50216/9657f7176cef2aa37e352d4933b2d2d5

And the output is `0.17964 0.00000 0.00000 0.82036`, which is close enough to the prediction on turnipprophet.io now (17.9% and 82.1%) that I'd say the current probability for this case is correct.

Issue #85, opened 2020-04-14
Spanish translation!


hicor13:
I'm in charge of an ACNH discord in Spanish and would love to help edit a Spanish version of the calculator, don't know much about HTML5 or coding in general, but could pass you a doc with full translations! Im awaiting any confirmation if my help is needed!

VHSCODE:
This is a good idea but as far as i know, all the text that the calculator uses is hardcoded, this could be fixed by adding multiple language support.

theRTC204:
There's definitely interest in supporting translations, but no decision has been made on the best way to go about it. There's more discussion about multi-language support going on in issue https://github.com/mikebryant/ac-nh-turnip-prices/issues/21.

theRTC204:
Forgot we had this issue and I made a new one #177, so I'll close this as a duplicate.

Issue #21, opened 2020-04-07
Translation


Adri1:
Hello!
It is possible in a near future to provide you a French translation? I would like to share your wonderful tool to some friends or people who are not good in English...

promarcel:
I can help with a german translation if help is needed :)

thetruegamer:
I'd be happy to do the french translation. I haven't done web for ages though, so don't know how you'd want to implement the language switch and such.

Adri1:
I don't now as well, I'm just a noob with a very few notions... I'm doing this request just to submit my own translation, but not the code behind.
But still, if i have to do any suggestion, I'd say it could be nice if the code recognize the laguage use by the navigator with and additional laguage switch in a corner of the page.

mikebryant:
I'm open to translations. I don't want to sink a vast amount of effort into setting up infra though, so I'm doing a bit of research into free translation platforms. If anyone knows of anything good, please let me know :)

promarcel:
@mikebryant There are many tools for manage translation or organize them. 
Unfortunately the most web-based services are paid plans and self-hosted solutions need some setup and infrastructure too.

The only free of charge alternative I know is https://poeditor.com
Alternatively we can just set up the strings inside the Repository here, for example by JSON.

For frontend rendering https://github.com/wikimedia/jquery.i18n can help (just a example, which can easily be integrated).

mikebryant:
I've raised a query with https://weblate.org/en-gb/ - see if we can get the free plan for libre projects :)

validvoid:
How about https://crowdin.com/ and I can help with a Chinese translation

NullYing:
Chinese translation with Wechat Mini Program

![image](https://user-images.githubusercontent.com/10618381/79094162-2dc38000-7d89-11ea-89e7-f498e24892ce.png)

mikebryant:
Can't use crowdin for open source unless the project has been active for 3 months

At least unless I pay for it :p

theRTC204:
Are we against maintaining our own translations? I get that it would be a ton more work, and it would be severly limited, but we could ask the community for translations as new regions start using the tool.

We could create the foundation of the localization by wiring up the language file, then it's just a matter of introducing new languages over time as we can get them.

Adri1:
I'm completely up to translate the whole website in French as soon as I know how we will process. From then, it will take me only around 15 to 30 minutes max to wrap up every entry to translate.
[...]

PR #135, opened 2020-04-16
feat: Taking a first stab at implementing translations

mikebryant:
Related to #21, #85

### DIFF
--- a/js/translations.js
+++ b/js/translations.js
@@ -1,0 +1,25 @@
+function updateContent() {
+  update();
+  $("body").localize();
+}
+
+i18next
+  .use(i18nextXHRBackend)
+  .use(i18nextBrowserLanguageDetector)
+  .init({
+    fallbackLng: 'en',
+    debug: true,
+    backend: {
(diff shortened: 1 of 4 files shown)


js/translations.js
```diff
@@ -0,0 +1,26 @@
...
+
+i18next
+  .use(i18nextXHRBackend)
+  .use(i18nextBrowserLanguageDetector)
+  .init({
+    fallbackLng: 'en',
+    debug: true,
+    backend: {
+      crossdomain: true,
+      loadPath: 'langs/{{lng}}.json',
```
theRTC204, line 14: Nitpick: I believe the common convention for most i18n implementations is to use a folder called `locales`. Makes no difference to the implementation, but could make it a bit more clear what the folder contains for people who deal with localization often.

mikebryant: I went with that from following: https://docs.weblate.org/en/latest/formats.html#js-i18next

But it looks like it's configurable so I'll rename it to locales. Thanks for the feedback :)

Issue #134, opened 2020-04-16
999 to 0


hunterxyz:
It looks like there is still a problem, those are my numbers, it brakes when I insert the 3rd number but the previous chart includes 61 as a possibility.

![image](https://user-images.githubusercontent.com/2712457/79458936-d9bbd400-7ff2-11ea-84f4-041a451598fa.png)

mikebryant:
That's really weird because those numbers work for me...

Just a thought, what happens if you force a refresh, ctrl+F5 etc?

![Screenshot from 2020-04-16 14-22-59](https://user-images.githubusercontent.com/1478103/79461242-ce19de80-7fed-11ea-9430-a58f8412d86c.png)

hunterxyz:
The refresh worked. I was sure I did it before opening this issue.

Sorry for the trouble.

Issue #133, opened 2020-04-16
Calculations on mobile


Jiopurr:
When having a startingbproce of 100 on a sunday the mobile version states that the max range for everyday is 999-0 and ignores the values you enter

mikebryant:
I can't reproduce I'm afraid - it's working for me. Can you provide a screenshot, or any console error output?

theRTC204:
I assume by "mobile version" they mean the PWA, in which case I suspect this may be a symptom of the service worker cache. My guess is, this person probably has a cached copy of the app when we had some broken code deployed.

@Jiopurr can you confirm whether you mean the installable app on an Android device, and if so, can you ensure your device is connected to a network and refresh the app page to make sure you get the latest code. Then, test again!

trombonekenny:
I'm seeing the same 999 to 0 bug on Firefox for MacOS as of this morning. Was working ok yesterday.

Console shows:
Failed to load ‘https://www.googletagmanager.com/gtag/js?id=UA-162470902-1’. A ServiceWorker passed a promise to FetchEvent.respondWith() that rejected with ‘no-match’.
Loading failed for the <script> with source “https://www.googletagmanager.com/gtag/js?id=UA-162470902-1”. turnipprophet.io:30:1
[PWA] Service Worker already found, skipping register turnipprophet.io:336:17
i18next::backendConnector: loading namespace translation for language en-US failed failed loading locales/en-US.json i18next.min.js:1:2589
i18next::backendConnector: loaded namespace translation for language en 
Object { patterns: {…}, times: {…} }
i18next.min.js:1:2589
i18next: languageChanged en-US i18next.min.js:1:2589
i18next: initialized 
Object { debug: true, initImmediate: true, ns: (1) […], defaultNS: (1) […], fallbackLng: (1) […], fallbackNS: false, whitelist: false, nonExplicitWhitelist: false, load: "all", preload: false, … }
i18next.min.js:1:2589
Array []
predictions.js:820:11

​

theRTC204:
I'm sorry, I am not able to reproduce what you are seeing here... Can you use the "Copy permalink" button to send us an example of what your inputs show?

trombonekenny:
https://turnipprophet.io?prices=93.88.87.97.75.72.68.64.....

![image](https://user-images.githubusercontent.com/10765199/79488439-56dd4e00-7fdf-11ea-9e66-c7842e740252.png)

trombonekenny:
[...]

PR #138, opened 2020-04-16
Fetch from internet first for service worker

peter50216:
This makes the page less likely to be in a broken state due to updated content.

Tested locally.


theRTC204:
I suspect that issues #133 and #134 might be caused by this, so probably good to get it out ASAP.

theRTC204:
You did it so ASAP that it happened before my comment came in! 😆

PR #143, opened 2020-04-16
Added missing i18n keys

promarcel:
As talked about I added the missing i18n keys to the source code inside the `index.html` and all JavaScript files which contain localized text.


locales/en.json
```diff
@@ -1,13 +1,80 @@
...
+  "general": {
+    "name": "Turnip Prophet",
+    "daisymae": "Daisy Mae"
+  },
+  "welcome": {
+    "salutation": "Hello, and welcome to the <b>Turnip Prophet</b> app on your Nook Phone.",
+    "description": "This app lets you track your island's turnip prices daily, but you'll have to put the prices in yourself!",
+    "conclusion": "After that, the Turnip Prophet app will <b>magically</b> predict the turnip prices you'll have for the rest of the week."
+  },
+  "firsttime": {
```
theRTC204, line 11: This key should probably be hyphonated as `first-time`.

locales/en.json
```diff
@@ -1,13 +1,80 @@
 {
+  "general": {
+    "name": "Turnip Prophet",
+    "daisymae": "Daisy Mae"
```
theRTC204, line 4: This key should probably be hyphonated as `daisy-mae`.

locales/en.json
```diff
@@ -1,13 +1,80 @@
...
   "times": {
     "morning": "AM",
     "afternoon": "PM"
+  },
+  "output": {
+    "title": "Output",
+    "chance": "% Chance",
+    "to": "to",
+    "minimum": "Guaranteed Minimum",
+    "maximum": "Guaranteed Maximum",
```
theRTC204, line 65: I believe the current language for the maximum field is actually `Potential Maximum` is it not?

locales/en.json
```diff
@@ -1,13 +1,80 @@
...
+  "weekdays": {
+    "monday": "Monday",
+    "tuesday": "Tuesday",
+    "wednesday": "Wednesday",
+    "thursday": "Thursday",
+    "friday": "Friday",
+    "saturday" : "Saturday",
+    "sunday": "Sunday",
+    "abr": {
+      "monday": "Mon ",
```
mikebryant, line 48: This is likely to go wrong when translators forget to do the extra space - no keys in the translation files should have leading or trailing spaces, we should do that in code if we're combining things

locales/en.json
```diff
@@ -1,13 +1,80 @@
...
+    "chart": {
+      "input": "Input Price",
+      "minimum": "Minimum",
+      "maximum": "Maximum"
+    }
+  },
+  "textbox": {
+    "description": "After you've listed some turnip prices, the Turnip Prophet will run some numbers and display the different possible patterns that your island may experience.",
+    "development": "This app is still in development, but will improve over time!",
+    "thanks": "None of this would have been possible without <a href=\"https://twitter.com/_Ninji/status/1244818665851289602?s=20\">Ninji's work</a> figuring out just how Timmy and Tommy value their turnips.",
```
mikebryant, line 75: I don't think we want html code in the translations, that's going to be harder for translators to get right

(Applies to several keys)

promarcel:
@theRTC204 Changes has been done.

locales/en.json
mikebryant: Ok, not sure how to actually do this better, so won't block anything on this - can refactor later

PR #146, opened 2020-04-16
Add payed price input

bake:
Hi, I've added an input for the price one payed for their turnips. This obviously does not impact the calculation, but it adds a horizontal line to the graph comparing it to the possible price range. Please let me know what you think about the idea.

### DIFF
--- a/index.html
+++ b/index.html
@@ -111,12 +111,19 @@
         </div>
 
         <div class="form__row">
+          <h6>Payed Price</h6>
+          <div class="input__group">
+            <label>What was the price you payed for the turnips this week? <i>(This value does not impact the calculation)</i></label>
+            <input type="number" id="payed" placeholder="..." />
+          </div>
+        </div>
+
+        <div class="form__row">
           <h6>Sunday</h6>
           <div class="input__group">
             <label>What was the price of turnips this week on your island? <i>(If this is your first time buying turnips, this field
                 will be disabled)</i></label>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -110,13 +110,20 @@ <h6>Previous Pattern</h6>
...
         </div>
 
+        <div class="form__row">
+          <h6>Payed Price</h6>
+          <div class="input__group">
+            <label>What was the price you payed for the turnips this week? <i>(This value does not impact the calculation)</i></label>
+            <input type="number" id="payed" placeholder="..." />
+          </div>
+        </div>
+
```
promarcel, line 120: This part is not needed because the "Payed Price" is the value from Sunday mentioned in the code down below so this part should be remove. Or am I wrong @theRTC204

bake: It might be a different number, depending on the island you bought the turnips.

theRTC204: You are correct @promarcel.

Further, this change flirts with the idea that it is a calculator for your profit... which is clearly detailed in the README as out of scope.

Ultimately @mikebryant will make the call though.

mikebryant:
Thanks for the contribution! I'm afraid I don't think this sits within the core mission of the tool, as a predictor for future prices. We need the buy price because it's part of the calculation, but I don't want to clutter the site with extra inputs that aren't needed to compute the prices or probabilities.

PR #147, opened 2020-04-16
Various UI improvements to chart

benru:
This PR has several small UI changes. If you'd prefer them separated out I can do that. Also this branch is live on: https://benru.github.io/ac-nh-turnip-prices/

I was focussing on improvements to the chart trying to stop it redrawing and animating between invalid states as you used it. The biggest change tries to prevent the chart from disappearing while the user is typing - e.g. if they are trying to type "123" but have only typed "1" (invalid in all patterns) then the chart would disappear and reappear. I added a change to detect invalid patterns (which should never happen with real data) and pause the redraw for 1 second in case the user is typing.

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -354,117 +354,88 @@ class Predictor {
 
     const buy_price = given_prices[0];
     let prob = 1;
-    let rate_range = [rate_min, rate_max];
+    let peak_range = [rate_min, rate_max];
+    let side_range = [rate_min, rate_max];
+    let min_pred = this.get_price(rate_min, buy_price);
+    let max_pred = this.get_price(rate_max, buy_price);
 
-    // * Calculate the probability first.
-    // Prob(middle_price)
-    const middle_price = given_prices[start + 1];
-    if (!isNaN(middle_price)) {
-      const min_pred = this.get_price(rate_min, buy_price);
-      const max_pred = this.get_price(rate_max, buy_price);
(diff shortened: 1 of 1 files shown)


benru:
Thank you both for your comments.

@mikebryant I think that "0.00%" does look better than "1.62e-7%" to non-technical users. In reality those patterns shouldn't be there and are only because of the fudge factor. If you remove that, they go away.

@theRTC204 I was initially trying to fix the graph not clearing when you pressed Reset, and was thinking about what the reset graph should show. The pregenerated graph is not dissimilar to the graph shown when you only have Sunday's price entered - I think it is interesting to see what a generic week looks like.

As for the debounce, I think that's what I've done except not at the input level it's after it's tried to update and realised that the new pattern is invalid. I was trying to keep the delay to a minimum when it does match patterns, but I could see that the behaviour might be surprising. Do you think I should switch to a debounce on all inputs regardless?

In general, since I will have to rebase this PR anyway, what do you think I should keep and what should be taken out?

mikebryant:
I think even without the fudging there will still be cases where there's a very unlikely pattern (when we hit boundary conditions)

When thinking about this further, given this is being special-cased in the code anyway, I'd prefer to show something like `<0.01%` - I think that's as useful to non-technical users as 0.00%, and satisfies my desire to also be correct

benru:
Apologies. I believe I have broken this PR because it shouldn't have been requesting a merge from my master branch, and now I've pushed something else there. I will take the feedback and make a smaller PR for the less controversial changes.

PR #160, opened 2020-04-17
Add Italian translation

THeK3nger:
Adding a candidate Italian translation. It seems fine to me in the layout (sometimes Italian may be a bit more wordy).


THeK3nger:
I should have fixed everything. 👍

PR #165, opened 2020-04-17
Update zh-CN.json

NullYing:
for more easier understand.

### DIFF
--- a/locales/zh-CN.json
+++ b/locales/zh-CN.json
@@ -22,8 +22,8 @@
     "decreasing": "递减型",
     "fluctuating": "波动型",
     "unknown": "不知道",
-    "large-spike": "大幅上涨型",
-    "small-spike": "小幅上涨型"
+    "large-spike": "大幅上涨（四期型）",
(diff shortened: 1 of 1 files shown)


locales/zh-CN.json
```diff
@@ -20,10 +20,10 @@
     "pattern": "趋势",
     "all": "所有趋势",
     "decreasing": "递减型",
-    "fluctuating": "波型",
+    "fluctuating": "波动型",
     "unknown": "不知道",
-    "large-spike": "三期型",
-    "small-spike": "四期型"
+    "large-spike": "大幅上涨（四期型）",
+    "small-spike": "小幅上涨（三期型）"
```
theRTC204, line 26: Would recommend dropping the `（四期型）` and `（三期型）` off these translations. Those suggest a numerical order ot the patterns that both does not map to the order the buttons appear on the website, and possibly worse they also don't match up to their in-game pattern number which would likely end up confusing people who are familiar with the extracted game code.

NullYing: Agree, I take new PR to drop it

Issue #136, opened 2020-04-16
UI Selector to manually change Language


mikebryant:
Current implementation will always choose the browser language

promarcel:
I will take care of this now.

mikebryant:
Closed by #168

Issue #41, opened 2020-04-08
Pattern Names


mikebryant:
I am currently thinking that having shorter pattern names would be desirable.

I also think having a more detailed description of how they would work be nice.

My thoughts as to how to solve this are to put them in the table with short names, and on hover or tap on mobile they should explain what happens in longer words.

mikebryant:
My proposed pattern names:
- Fluctuating
- Small spike
- Decreasing
- Large spike

tylermatchett:
@theRTC204 and I were just discussing that on #34.
I agree, I like the shorter names for readability/quick user parsing, and a hover over or legend to break it down to still have the relation to the game code.

theRTC204:
How do the proposed names map to the actual patterns? Would the following mapping be what you are thinking?

PATTERN 0: `high, decreasing, high, decreasing, high` -> `Fluctuating`
PATTERN 1: `decreasing middle, high spike, random low` -> `Large spike`
PATTERN 2: `consistently decreasing` -> `Decreasing`
PATTERN 3: `decreasing, spike, decreasing` -> `Small spike`

mikebryant:
Yup, that's exactly right

theRTC204:
I like this idea! I think there is a lot of user confusion caused by the names from the game, so having simpler names and more clear descriptions of their behavior seems like a good idea.

It might be of some value to at least keep the names from the games somewhere, for people who may be using this tool along with other tools that are based on Ninji's extracted code.

tylermatchett:
should we include them in a data-attribute to keep the relation? would make it easy to parse for others if they need to.

Aliervo:
I like the proposed names. I've been looking through the code trying to troubleshoot https://github.com/mikebryant/ac-nh-turnip-prices/issues/40, but I keep getting a little lost with the data-mined names.

theRTC204:
This issue can probably be closed @mikebryant, since the names have already been changed in https://github.com/mikebryant/ac-nh-turnip-prices/pull/42.

mikebryant:
Names, yes, but I still want too add more detailed descriptions in a tooltip type thing

Jollus174:
I like this idea of presenting more information about different types of fluctuations, although placing it inside a tooltip feels a bit like obscuring the UX - the info will only be revealed for one tooltip at a time as the user hovers over a radio button.

Something like this could be feasible, where the information is presented beneath each button. although translations would need to be added for each of the already provided languages.
[...]

PR #168, opened 2020-04-17
Add a simple language selector. 

DevSplash:
https://github.com/mikebryant/ac-nh-turnip-prices/issues/136
The selector at the bottom of the page.

### DIFF
--- a/js/translations.js
+++ b/js/translations.js
@@ -4,22 +4,40 @@ function updateContent() {
 }
 
 i18next
-  .use(i18nextXHRBackend)
-  .use(i18nextBrowserLanguageDetector)
-  .init({
-    fallbackLng: 'en',
-    debug: true,
-    backend: {
-      loadPath: 'locales/{{lng}}.json',
-    },
-  }, function(err, t) {
-    jqueryI18next.init(i18next, $);
-    i18next.on('languageChanged', () => {
-      updateContent();
-    });
-
-    // init set content
-    $(document).ready(initialize);
-    $(document).on("input", updateContent);
-    $('input[type = radio]').on("change", updateContent);
+.use(i18nextXHRBackend)
+.use(i18nextBrowserLanguageDetector)
+.init({
+  fallbackLng: 'en',
+  debug: true,
+  backend: {
+    loadPath: 'locales/{{lng}}.json',
+  },
+}, (err, t) => {
+  const languages = [
+    ["en", "English"],
+    ["zh-CN", "简体中文"],
+    ["zh-TW", "繁體中文"],
+  ],
+  languageSelector = $('#language');
+  languages.map(([code, name]) => {
+    languageSelector.append(`<option value="${code}"${code == i18next.language ? ' selected' : ''}>${name}</option>`);
   });
+  languageSelector.on('change', function () {
+    if (this.value == i18next.language)
+      return;
+    i18next.changeLanguage(this.value);
+  });
+  jqueryI18next.init(i18next, $);
+  i18next.on('languageChanged', lng => {
+    if (!languageSelector.find(`[value=${lng}]`).length) {
+      i18next.changeLanguage('en');
+      return;
+    }
+    languageSelector.val(lng);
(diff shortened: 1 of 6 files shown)


js/translations.js
```diff
@@ -1,25 +1,43 @@
...
+.use(i18nextXHRBackend)
+.use(i18nextBrowserLanguageDetector)
+.init({
+  fallbackLng: 'en',
+  debug: true,
+  backend: {
+    loadPath: 'locales/{{lng}}.json',
+  },
+}, (err, t) => {
+  const languages = [
```
promarcel, line 16: Can you please directly add the following line: `['de', 'Deutsch'],`
Thank you very much!

DevSplash: Done.

locales/zh-CN.json
```diff
@@ -22,8 +22,8 @@
     "decreasing": "递减型",
     "fluctuating": "波动型",
     "unknown": "不知道",
-    "large-spike": "大幅上涨（四期型）",
-    "small-spike": "小幅上涨（三期型）"
+    "large-spike": "大幅上涨（三期型）",
+    "small-spike": "小幅上涨（四期型）"
```
theRTC204, line 26: I'm not fond of the pattern description in brackets here... `（三期型）`  and `（四期型）`, as it is not a direct translation of the text - it is additional context that is not provided to other languages. I would prefer to see this removed as there is additional conversation taking place in #41 about how to better communicate the behavior of the patterns.

DevSplash: I added a demo image. 
As for `（三期型）` and `（四期型）`, you can find the reason they are so named [here](https://github.com/mikebryant/ac-nh-turnip-prices/pull/166). 
> (The "small-spike" is called "四期型" because the price would increase four (四) times in the week, and "large-spike" == "三期型" because the price would increase three (三) times in the week.)

But I have no idea whether it should be deleted. I thought @peter50216 could offer some ideas.

theRTC204:
The selector could use some styling love, but we can raise that in another issue if that's outside your wheelhouse.

DevSplash:
> The selector could use some styling love, but we can raise that in another issue if that's outside your wheelhouse.

OK, I’ll try it.

promarcel:
From the UX side we should investigate as @theRTC204 mentioned.

I think we need to move the language selector also to the top of the page that people directly can see that they can switch the language.

theRTC204:
I thought about suggesting moving it, but then since the app automatically changes to your system language, I felt being at the bottom was ok.

js/translations.js
```diff
@@ -1,25 +1,46 @@
...
+    ['de', 'Deutsch'],
+    ['en', 'English'],
+    ['zh-CN', '简体中文'],
+    ['zh-TW', '繁體中文']
+  ],
```
mikebryant, line 21: Should this be `;`?

locales/zh-CN.json
```diff
@@ -22,8 +22,8 @@
     "decreasing": "递减型",
     "fluctuating": "波动型",
     "unknown": "不知道",
-    "large-spike": "大幅上涨（四期型）",
-    "small-spike": "小幅上涨（三期型）"
```
mikebryant, line 26: This feels like an unrelated change - can it be removed?

mikebryant: (Also other changes like this - I'd like to have this just be a PR about the language selector, not other things)

js/translations.js
DevSplash: The `languageSelector` is also defined, so it is correct. 
```suggestion
  const languages = [
    ['de', 'Deutsch'],
    ['en', 'English'],
    ['zh-CN', '简体中文'],
    ['zh-TW', '繁體中文']
  ],
  languageSelector = $('#language');
```

locales/zh-CN.json
DevSplash: I'll remove them and put up a new PR for them later.

PR #169, opened 2020-04-17
Added German translation

promarcel:
Added German translation files.

Changes inside the `styles.css` need to be applied because the cells are to tight for the German words. 😂


mikebryant:
Thanks for adding this!

PR #172, opened 2020-04-17
dropping 三期型、四期型

NullYing:
https://github.com/mikebryant/ac-nh-turnip-prices/pull/165#discussion_r410357473

> Would recommend dropping the `（四期型）` and `（三期型）` off these translations. Those suggest a numerical order ot the patterns that both does not map to the order the buttons appear on the website, and possibly worse they also don't match up to their in-game pattern number which would likely end up confusing people who are familiar with the extracted game code.

### DIFF
--- a/locales/zh-CN.json
+++ b/locales/zh-CN.json
@@ -22,8 +22,8 @@
     "decreasing": "递减型",
     "fluctuating": "波动型",
     "unknown": "不知道",
-    "large-spike": "大幅上涨（三期型）",
-    "small-spike": "小幅上涨（四期型）"
+    "large-spike": "大幅上涨",
+    "small-spike": "小幅上涨"
   },
   "prices": {
     "description": "本周你的岛上大头菜的购买价格是多少？<i>（如果你是第一次购买大头菜，这个字段不可用）</i>",
@@ -42,7 +42,7 @@
     "wednesday": "周三",
     "thursday": "周四",
     "friday": "周五",
-    "saturday" : "周六",
+    "saturday": "周六",
     "sunday": "周日",
     "abr": {
       "monday": "周一",
@@ -50,7 +50,7 @@
       "wednesday": "周三",
       "thursday": "周四",
       "friday": "周五",
-      "saturday" : "周六"
+      "saturday": "周六"
     }
   },
   "times": {
--- a/locales/zh-TW.json
+++ b/locales/zh-TW.json
@@ -20,10 +20,10 @@
     "pattern": "模型",
     "all": "所有模型",
     "decreasing": "遞減型",
-    "fluctuating": "波型",
+    "fluctuating": "波動型",
     "unknown": "不知道",
-    "large-spike": "三期型",
-    "small-spike": "四期型"
+    "large-spike": "大幅上漲",
+    "small-spike": "小幅上漲"
   },
   "prices": {
     "description": "本週大頭菜的購買價格是多少？<i>(如果這是您第一次購買大頭菜，則此欄位會被停用)</i>",
@@ -42,7 +42,7 @@
     "wednesday": "星期三",
     "thursday": "星期四",
     "friday": "星期五",
-    "saturday" : "星期六",
+    "saturday": "星期六",
     "sunday": "星期日",
     "abr": {
       "monday": "週一",
@@ -50,7 +50,7 @@
       "wednesday": "週三",
       "thursday": "週四",
       "friday": "週五",
-      "saturday" : "週六"
+      "saturday": "週六"
     }
   },
   "times": {


locales/zh-CN.json
```diff
@@ -22,8 +22,8 @@
     "decreasing": "递减型",
     "fluctuating": "波动型",
     "unknown": "不知道",
-    "large-spike": "大幅上涨（三期型）",
-    "small-spike": "小幅上涨（四期型）"
+    "large-spike": "大幅上涨",
+    "small-spike": "小幅上涨"
```
DevSplash, line 26:
```suggestion
    "large-spike": "三期型",
    "small-spike": "四期型"
```
In fact, I think it would be better to use `三期型` and `四期型` (Remove `大幅上涨` and `小幅上涨`).

locales/zh-TW.json
```diff
@@ -20,10 +20,10 @@
     "pattern": "模型",
     "all": "所有模型",
     "decreasing": "遞減型",
-    "fluctuating": "波型",
+    "fluctuating": "波動型",
     "unknown": "不知道",
-    "large-spike": "三期型",
-    "small-spike": "四期型"
+    "large-spike": "大幅上漲",
+    "small-spike": "小幅上漲"
```
DevSplash, line 26:
```suggestion
    "large-spike": "三期型",
    "small-spike": "四期型"
```

locales/zh-CN.json
NullYing: 无法理解啥是三期型四期型，可以麻烦解释一下？对于普通人来说，跟我一样有疑问的会有更多

DevSplash: > 无法理解啥是三期型四期型，可以麻烦解释一下？对于普通人来说，跟我一样有疑问的会有更多

简单来说，三期型上涨三次，四期型上涨四次。
https://github.com/mikebryant/ac-nh-turnip-prices/pull/166

NullYing: 那么我觉得可以翻译成这样，而不是简单的三期型和四期型

三期型（上涨三次）
四期型（上涨四次）

NullYing: 不经过解释的话，很多人都不会知道这两词的意思

DevSplash: 因此在最初翻译的时候我就考虑了很久，我也并不确定哪种翻译才是最佳的。

NullYing: 三期型和四期型最开始再繁体的资料里面看到。如果当做资料来看的话，会有更多解释来支撑这个翻译，而这个网站一般针对更多的是普通玩家，不建议沿用

NullYing: 已修正为，让更多人容易理解

三期型（上涨三次）
四期型（上涨四次）

theRTC204: My worry about "up three times" or "up four times" is they don't explain that one is better than the other. People will need to learn what they mean from somewhere else, which is not great.

Three large increases could be better than four small increases.

A way to explain that one has a large increase, and the other has a small increase seems easier for regular people to understand.

That is why in English we chose "large spike" and "small spike"

NullYing: It maybe better? It's too difficult

三期型（大幅上涨三次）
四期型（小幅上涨四次）

NullYing: 大幅上涨三次
小幅上涨四次

DevSplash: > My worry about "up three times" or "up four times" is they don't explain that one is better than the other. People will need to learn what they mean from somewhere else, which is not great.
> 
> Three large increases could be better than four small increases.
> 
> A way to explain that one has a large increase, and the other has a small increase seems easier for regular people to understand.
> 
> That is why in English we chose "large spike" and "small spike"

The concept of large and small is not clear. Perhaps it would be more intuitive to let  users judge by the number of rises.

DevSplash: > It maybe better? It's too difficult
> 
> 大幅上涨三次
> 小幅上涨四次

三期型（大幅上涨三次）
四期型（小幅上涨四次）

NullYing: > > It maybe better? It's too difficult
> > 大幅上涨三次
> > 小幅上涨四次
> 
> 三期型（大幅上涨三次）
> 四期型（小幅上涨四次）

It's too many word

DevSplash: Perhaps it would be better to explain the patterns by adding title attribute.

NullYing: I think 大幅上涨三次 and 小幅上涨四次 is enough to explain, and it seems easier for regular people to understand.

peter50216:
There was a discussion about this on discord yesterday, and I think the conclusion was that we can use the "三期型" "四期型" as pattern names since they're somewhat commonly referred pattern names in many Chinese articles about turnips, and rely on #41 to have a proper solution on explaining the pattern meaning clearly.

locales/zh-CN.json
theRTC204: There is an issue open #41 for adding tooltip/hover that explains how the pattern works.

The idea behind the title is to give a simple name that explains that overall result of the pattern, not each phase of that pattern.

For example, "large spike" means the overall result is a big increase in sell price, but doesn't indicate how many periods the increase will last.

theRTC204:
Yes, ideally we would use a term that is common in the native language as a simple title of the pattern, and then when we add the tooltips that better explain the behavior of the patterns we will localize that information as well.

NullYing:
> Yes, ideally we would use a term that is common in the native language as a simple title of the pattern, and then when we add the tooltips that better explain the behavior of the patterns we will localize that information as well.

"三期型" "四期型" is’t common in the native language as a simple, He just appeared in an analyzed article only. After all, people have preconceived habits.

mikebryant:
What's the status of this? I don't understand the language or the discussion here, so I'm a bit in the dark :p

locales/zh-CN.json
```diff
@@ -77,4 +77,4 @@
     "contributors-text": "哦！别忘记感谢那些至今为止做出过贡献的人。",
     "contributors": "贡献者"
   }
-}
+}
```
peter50216, line 80: nit: should have new line at end of file.

theRTC204:
I don't speak the language either, but Google seems to agree with @peter50216 in that it suggests the translation is "soars three times". If that is actually the translation, then it is incorrect.

KKKKaiZhu:
"大幅上涨" / "小幅上涨" makes more sense to me, and I think it will be more intuitive to most people.
Don't think many player (including me!) have a good understanding to "三期型" / "四期型".
"大幅上漲三次" and "小幅上漲四次" are inaccurate because the pattern doesn't actually "soars 3 times" or "drops 4 times".

yxwangcs:
> Don't think many player (including me!) have a good understanding to "三期型" / "四期型".

I second that. In fact, I believe we should assume the opposite, especially for the new players that just started AC:NH.

My personal preference is that we should make the name as simple as possible (therefore "大幅上涨" / "小幅上涨"), and let the tooltips do everything else.

PR #183, opened 2020-04-18
Add pt-BR translation

arctco:
I managed to translate the website to Brazilian Portuguese, I hope this helps anyone who wants to use the website!


arctco:
Done! Sorry I didn't know about the `js/translations.js` file, it should be all good now, if there is anything else, please let me know!

PR #185, opened 2020-04-19
Russian translation for the app

captaincurls:
Hello! I made a Russian translation for this project, would appreciate it being added to the app :)


captaincurls:
Updated the selector!

js/translations.js
```diff
@@ -16,8 +16,11 @@ i18next
   const languages = [
     ['de', 'Deutsch'],
     ['en', 'English'],
+    ['fr', 'Français'],
```
mikebryant, line 19: This and the Portuguese line shouldn't be changing in this PR - Not sure what happened here, I think you need to `git rebase` to remove conflicts

captaincurls: Sorry for the confusion, it's my first time contributing. I will upload a file with only one line added shortly!

mikebryant: No problem :)

Issue #124, opened 2020-04-15
Feature request: highlight the cell with the best price in each row


zgerlock:
While it's not too hard to parse through them it would be nice if the peaks were highlighted since that's the main thing you care about. Would also serve to visually break up what is otherwise a huge grid of numbers. You could also highlight days of guaranteed profit one color and the peak another, that seems even better.

ps. thanks all for putting this together, it's dope. Love the nook phone styling

mikebryant:
We now have some colours!

I think that satisfies this issue

And thanks for your comments. Glad you're liking the app :)

PR #189, opened 2020-04-19
Classes in table

abrad45:
Adds some class names and `[data-min]` and `[data-max]` attributes to support styling in the future.

This adds no styling whatsoever, but seeks to lay the groundwork that can be used to solve some existing, open issues such as #12, #124, etc.

I didn't want to style now due to the design and accessibility considerations that should be taken to ensure the best results.


abrad45:
Screen Shot 2020-04-19 at 12 15 26 PM

PR #190, opened 2020-04-19
Update ru.json

captaincurls:
I'm so sorry I noticed a few typos in the beginning right after the site went live :( I'm not sure I should be opening a new request for this...


mikebryant:
Pull requests are appropriate for all updates :)

I have nothing against small ones :)

PR #196, opened 2020-04-20
Sort language list alphabetically. 

DevSplash:


DevSplash:
> This is probably fine for now, but long term we'll likely need to apply a sorting filter on this array as enforcing this through PRs will be troublesome.

It's easy, but I recommend still making sure the language list in script is alphabetized. 
`const languages = [(language list)].sort();`

PR #219, opened 2020-04-21
fix: Only use a fudge factor if we fail to find patterns normally

mikebryant:


mmajko:
It seems, that this change made my predictions very different, is it on purpose?

Before merging this, for this data input:
https://turnipprophet.io?prices=103.81.77.68.145........

I had predictions ranging between 50-207, now the predictions are rather capped to 52-145:
Screenshot 2020-04-21 at 18 50 07

PR #243, opened 2020-04-23
Use exact numbers for steady state probabilities

mcpower:
See #90 for previous discussion. This uses the exact probabilities instead of the rounded decimals (modulo floating point error).

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -787,8 +787,12 @@ class Predictor {
 
   get_transition_probability(previous_pattern) {
     if (typeof previous_pattern === 'undefined' || Number.isNaN(previous_pattern) || previous_pattern === null || previous_pattern < 0 || previous_pattern > 3) {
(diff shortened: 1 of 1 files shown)


js/predictions.js
```diff
@@ -787,8 +787,7 @@ class Predictor {
 
   get_transition_probability(previous_pattern) {
     if (typeof previous_pattern === 'undefined' || Number.isNaN(previous_pattern) || previous_pattern === null || previous_pattern < 0 || previous_pattern > 3) {
-      // TODO: Fill the steady state pattern (https://github.com/mikebryant/ac-nh-turnip-prices/pull/90) here.
-      return [0.346278, 0.247363, 0.147607, 0.258752];
+      return [4530/13082, 3236/13082, 1931/13082, 3385/13082];
```
peter50216, line 790: I'd prefer leaving a comment here saying this is steady state probability and have a link to the issue.

mcpower: Done.

PR #245, opened 2020-04-23
Add Hungarian translation

smileyhead:
Hello there.

### DIFF
--- a/locales/hu.json
+++ b/locales/hu.json
@@ -69,7 +69,7 @@
     }
   },
   "textbox": {
-    "description": "Miután megadtál néhány árat, a Retekpróféta kiszámolja és megjeleníti az összes lehetséges mintát ami a héten megjelenhet.",
+    "description": "Miután megadtál néhány árat, a Turnip Prophet kiszámolja és megjeleníti az összes lehetséges mintát ami a héten megjelenhet.",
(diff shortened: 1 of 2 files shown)


smileyhead:
Like this?

locales/hu.json
```diff
@@ -0,0 +1,80 @@
+{
+  "general": {
+    "daisy-mae": "Daisy Mae"
+  },
+  "welcome": {
+    "salutation": "Üdvözöllek a Nookfonod <b>Retekpróféta</b> alkalmazásában!",
```
theRTC204, line 6: Please leave uses of the name "Turnip Prophet" without translation. We are treating it as a proper name which would not be translated.

locales/hu.json
```diff
@@ -0,0 +1,80 @@
...
+    "minimum": "Garantált minimum",
+    "maximum": "Lehetséges maximum",
+    "chart": {
+      "input": "Megadott ár",
+      "minimum": "Garantált minimum",
+      "maximum": "Lehetséges maximum"
+    }
+  },
+  "textbox": {
+    "description": "Miután megadtál néhány árat, a Retekpróféta kiszámolja és megjeleníti az összes lehetséges mintát ami a héten megjelenhet.",
```
mikebryant, line 72: App name needs to be kept here too :)

smileyhead:
Oh, sorry about that. Fixed.

PR #249, opened 2020-04-23
Update README.md

promarcel:
I improved the README.md file added some styling and much more information.

Edit: Preview: https://github.com/promarcel/ac-nh-turnip-prices/blob/update-readme/README.md

PR #252, opened 2020-04-23
Update readme

### LINKED ISSUE #178: Text does not load when running from local filesystem
None of the text gets loaded when running from the local filesystem.

Looking at the console this seems to be the error

`Access to XMLHttpRequest at 'file:///D:/Download/ac-nh-turnip-prices-master/ac-nh-turnip-prices-master/locales/en.json' from origin 'null' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https.`

Lysoun:
I have added the fixes that was requested in #249, the original pull request

Closes #178

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,34 +24,12 @@ For details don't hesitate to talk to us, e.g. using our Discord server.
 
 At first please have a look at our current project scope:
 
-<table width="100%">
-	<thead>
-		<tr>
-			<th>Turnip Prophet is for</th>
-			<th>Turnip Prophet is <u>not</u> for</th>
-		</tr>
-	</thead>
-	<tbody>
-		<tr>
-			<td>
-				<ul>
-					<li>A predictor for future prices that week</li>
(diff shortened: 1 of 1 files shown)


Lysoun:
I could not find a way to center the picture so I removed it... If you find a way to do it using markdown, that would be appreciated.

theRTC204:
Markdown doesn't support positioning in that way. You could leave the image in place, but it would be left aligned always.

Lysoun:
I am pretty sure you can squash them on merge, can't you?

theRTC204:
@Lysoun I can, but a maintainer really shouldn't have to cleanup a contribution.

Lysoun:
But using the github squash makes it so much easier and it even properly adds the co-authors in the squash commit while I would have to ask @promarcel their email to properly push a squashed commit with the co-author. It is just two clicks away from the "merge" button in github

Issue #258, opened 2020-04-24
broken price predictions when the sunday price isn't specified


Dzwiedziasty:
https://turnipprophet.io?prices=.82.78.74.69.66.62.54.95.142.560..&pattern=3

setting the sunday price to around 100 fixes the graph

mikebryant:
Fixed by #262 :)

@Dzwiedziasty thanks for reporting this. Best of luck with your turnip prices!

Dzwiedziasty:
#262 didn't fix the issue fully - the large spike pattern still isn't predicted correctly if the last price isn't specified
https://turnipprophet.io?prices=.82.78.74.69.66.62.54.95.142...&pattern=3

PR #262, opened 2020-04-24
Calculate with estimated buy price without modifying original input values

FrederikGoovaerts:
I think this fixes Issue #258.

I went through the code to get a grasp on it and while hunting for this bug, saw that when looping into this function multiple times, the value 110 was still present in the sell-prices array, not triggering the `isNaN` check on subsequent iterations.

Please review thoroughly, I'm not sure what I might break by NOT modifying this array, but to my understanding, this is the actual fix.

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -810,13 +810,14 @@ class Predictor {
   * generate_possibilities(sell_prices, first_buy, previous_pattern) {
     if (first_buy || isNaN(sell_prices[0])) {
       for (var buy_price = 90; buy_price <= 110; buy_price++) {
-        sell_prices[0] = sell_prices[1] = buy_price;
+        const temp_sell_prices= sell_prices.slice()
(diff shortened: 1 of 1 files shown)


js/predictions.js
```diff
@@ -810,13 +810,14 @@ class Predictor {
   * generate_possibilities(sell_prices, first_buy, previous_pattern) {
     if (first_buy || isNaN(sell_prices[0])) {
       for (var buy_price = 90; buy_price <= 110; buy_price++) {
-        sell_prices[0] = sell_prices[1] = buy_price;
+        const temp_sell_prices= sell_prices.slice()
```
peter50216, line 813: nit: space before =, and ; at end of line.

FrederikGoovaerts: Dang, I've become so used to Prettier auto-kicking in that I've become lazy with typing my code in a decent way... I had it disabled for this project because it goes berserk on some of these files with the default settings.

Anyway, fixed!

PR #264, opened 2020-04-25
feat: Add link to GitHub sponsors button

### LINKED ISSUE #141: Patreon or donations link
Ability to give donations to the developers of the tool.

mikebryant:
Closes #141


mikebryant:
> My only question is about i18n. Since I haven't tested this myself, does the library fall back to the default on a per-key basis? Or will this line not appear in any other language until their locale file is updated?

It falls back on a per-key basis. So that line will show up in English until it gets translated.

Issue #269, opened 2020-04-25
Table Overflows off side until a sell price entered.


Bpendragon:
(currently) offending pattern: https://turnipprophet.io?prices=90 (note, initial price chosen is 2 digits to give system best chance of not overflowing)
screencap: 
![image](https://user-images.githubusercontent.com/5448822/80285960-6e05f500-86dd-11ea-9bb0-f0adb78f38a9.png)

As soon as any valid price is entered for Monday the overflow disappears: https://turnipprophet.io?prices=100.110........... (note, chose a 3 digit Sunday value value valid 3-digit Mon AM value for for Fluctuating pattern to give worst chance at overflow, works with valid value in any slot whatsoever, just choose Mon AM for simplicity)

Scrrencap: 
![image](https://user-images.githubusercontent.com/5448822/80286014-c76e2400-86dd-11ea-9b63-70d9470f4dab.png)

This occurs on latest: Chrome, Firefox, and Edge, at viewports from ~1400px across (and smaller and the table also gains a vertical scrollbar), all the way up to 2560px across (that is, my 1440p monitor)

Bpendragon:
Further investigation and possible solutions discussed over on my (admittedly hasty) PR:

https://github.com/mikebryant/ac-nh-turnip-prices/pull/271#issuecomment-619451959

mikebryant:
We've actually cut the Sunday column now, which makes the overall size smaller. Is this still an issue for you?

Bpendragon:
> We've actually cut the Sunday column now, which makes the overall size smaller. Is this still an issue for you?

This fixed the issue. As discussed in the linked PR, only needed about 40 px of extra room, and removing that column was more than enough. 

Closing this issue, and will close related PR momentarily.

PR #271, opened 2020-04-25
Shrink L/R padding on turnip table to fix overflow

Bpendragon:
Prevents overflow on desktop early in the week when only the Sunday price is known.

This is the simplest fix to #269 I could find. 

Discussion and alternate options encouraged.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -382,7 +382,7 @@ input[type=number] {
 #turnipTable td {
   white-space: nowrap;
   max-width: 150px;
-  padding: 6px 4px;
+  padding: 6px 2px;
   text-align: center;
   border-right: 1px solid rgba(0, 0, 0, 0.03);
   border-bottom: 1px solid rgba(0, 0, 0, 0.15);


theRTC204:
I don't know if I like the idea of shrinking the padding. We had a PR not too long ago to _increase_ the padding because some people found it hard to read the table with everything so squished.

Bpendragon:
> I don't know if I like the idea of shrinking the padding. We had a PR not too long ago to _increase_ the padding because some people found it hard to read the table with everything so squished.

That's completely fair. I'll try some other things out, obviously it's a very small error and I don't want to break things or accessibility to fix it

Bpendragon:
Further investigation shows that in the worst case scenario (112+ buy price, largest glyphs)  the `.nook-phone` element needs to increase by `41px`, or the content of the `#turnipTable` needs to decrease by a proportional amount. 

a `40px` increase to `.nook-phone` does remove the overflow and left/right scrollbar from the table, but on hover over any table element a vertical scrollbar appears that has 1px of travel.

Possible solutions that I've explored:

* Decrease `turnipTable td` L/R padding to 2px (originally submitted solution)
* Decrease `turnipTable td` L/R padding to 3px AND increase `.table-wrapper` max width from 98% to 99% (creates a slightly lopsided look in the early part of the week
* Decrease `turnipTable` `font-size` to `.9rem` (bad idea, shrinks text, hurting readability
* Increase `.nook-phone` `max-width` to 1441px
* Extra JS that shrinks some things when no sell prices filled out (have not investigated deeply)

Bpendragon:
Closing dues to the fact that Sunday column is now removed from chart, giving sufficient space to prevent the scrollbar from appearing.

PR #276, opened 2020-04-26
fix: prevent table being truncated in Chrome

### LINKED ISSUE #274: Table not rendered correctly on Android device (Chrome)
- Link: https://turnipprophet.io?prices=93&pattern=2
- Device: Android 8.0.0
- Browser: Chrome: 81.0.4044.111

The table is truncated if there are too many patterns to show. I think it's Chrome's bug because it looks great in Firefox (same device).

Screenshot_20200426-172050

### LINKED ISSUE #322: Invalid table viewing on Mi9 Lite
Mi9 Lite
Android 10
Google Chrome 81.0.4044.117

After adding a price, I can't see full table of price, only third of them

Screenshot_2020-05-03-14-47-36-347_com android chrome

akccakcctw:
Fixes #274
Fixes #322 

I just try my best not to change anything else, tested in Chrome and Firefox (Android).


css/styles.css
```diff
@@ -328,7 +328,6 @@ input[type=number] {
   padding: 16px;
   margin: 0px auto;
   box-sizing: border-box;
-  overflow-x: auto;
   border-radius: 2px;
```
boris-hocde, line 331: @akccakcctw I think that this bug is caused by `.table-wrapper { border-radius: 2px; }`, which doesn't seem to make any visible change. What about getting rid of this rule rather than changing the table display? 
```suggestion
```

akccakcctw:
@boris-hocde Thanks, I don't know why but remove `border-radius` fix this bug!

PR #277, opened 2020-04-26
Various improvements to PDF

mcpower:
- Use [x, x+1) instead of [x-0.5, x+0.5), ensuring that min_value() and max_value() return integers
- Fix range_of and max_value being off by epsilon
- Use improved Kahan–Babuska algorithm for floating point summation

These improvements:
- improve the accuracy of PDF by at least a factor of 2 when calculating the Irwin-Hall distribution with n=3
- fixes pdf.range_limit([pdf.min_value(), pdf.max_value()]) incorrectly mutating the PDF and returning a value that is not 1
- fixes the decreasing pattern with buy=100 incorrectly being calculated with max predictions which are 1 too high

The code used for testing the accuracy of this PDF implementation is shown below. `test_pdf_1` uses the [Irwin hall distribution](https://en.wikipedia.org/wiki/Irwin%E2%80%93Hall_distribution) with n=3, as it's a non-trivial distribution created by the summation of uniform distributions with a closed form.

```javascript
function float_sum(input) {
  let sum = 0;
  let c = 0;
  for (let i = 0; i < input.length; i++) {
    const cur = input[i];
    const t = sum + cur;
    if (Math.abs(sum) >= Math.abs(cur)) {
      c += (sum - t) + cur;
    } else {
      c += (cur - t) + sum;
    }
    sum = t;
  }
  return sum + c;
}
// These are all off by a factor of 6.
function irwin_hall_3_integral_0(x) {
  return [x*x*x];
}
function irwin_hall_3_integral_1(x) {
  return [-x*x*x*2, 9*x*x, -9*x];
}
function irwin_hall_3_integral_2(x) {
  return [x*x*x, -9*x*x, 27*x];
}
function irwin_hall_3_cdf(x) {
  let out = [];
  if (x > 0) {
      out.push(...irwin_hall_3_integral_0(Math.min(x, 1)));
      out.push(...irwin_hall_3_integral_0(0).map(x => -x));
  }
  if (x > 1) {
      out.push(...irwin_hall_3_integral_1(Math.min(x, 2)));
      out.push(...irwin_hall_3_integral_1(1).map(x => -x));
  }
  if (x > 2) {
      out.push(...irwin_hall_3_integral_2(Math.min(x, 3)));
      out.push(...irwin_hall_3_integral_2(2).map(x => -x));
  }
  return out;
}
// This is NOT off by a factor of 6.
function irwin_hall_3_range(l, r) {
  return float_sum([...irwin_hall_3_cdf(r), ...irwin_hall_3_cdf(l).map(x => -x)]) / 6;
}

// old n=1000:
//   max: 4.161661899424948e-10
//   rms: 2.9422043359977044e-10
// new n=1000:
//   max: 1.666680532735193e-10
//   rms: 1.178511301945725e-10
//
// old n=10000:
//   max: 3.677530149978278e-13
//   rms: 2.73381830629247e-13
// new n=10000:
//   max: 1.6812948738309866e-13
//   rms: 1.1785321393575006e-13
function test_pdf_1(n=1000) {
  const pdf = new PDF(0, n);
  pdf.decay(-n, 0);
  pdf.decay(-n, 0);
  const errors = pdf.prob.map((val, i) => Math.abs(val - irwin_hall_3_range(...pdf.range_of(i).map(x => x/n))));
  console.log("max: " + Math.max(...errors));
  console.log("rms: " + Math.sqrt(float_sum(errors.map(x => x*x)) / errors.length));
}

// old: [0.9994999980010354, 0.9997498739377202, 0.9998749051797337]
// new: [1, 1, 1]
function test_pdf_2(l=0, r=1000) {
  const pdf = new PDF(l, r);
  // These should all be 1.
  console.log([pdf.range_limit([l, r]), pdf.range_limit([l, r]), pdf.range_limit([l, r])]);
}
```

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -56,10 +56,64 @@ function range_intersect_length(range1, range2) {
   return range_length(range_intersect(range1, range2));
 }
 
+/**
+ * Accurately sums a list of floating point numbers.
+ * See https://en.wikipedia.org/wiki/Kahan_summation_algorithm#Further_enhancements
+ * for more information.
+ * @param {number[]} input
+ * @returns {number} The sum of the input.
+ */
+function float_sum(input) {
+  // Uses the improved Kahan–Babuska algorithm introduced by Neumaier.
+  let sum = 0;
+  // The "lost bits" of sum.
+  let c = 0;
+  for (let i = 0; i < input.length; i++) {
+    const cur = input[i];
+    const t = sum + cur;
+    if (Math.abs(sum) >= Math.abs(cur)) {
+      c += (sum - t) + cur;
+    } else {
+      c += (cur - t) + sum;
+    }
+    sum = t;
+  }
+  return sum + c;
+}
+
+/**
+ * Accurately returns the prefix sum of a list of floating point numbers.
+ * See https://en.wikipedia.org/wiki/Kahan_summation_algorithm#Further_enhancements
+ * for more information.
+ * @param {number[]} input
+ * @returns {[number, number][]} The prefix sum of the input, such that
+ * output[i] = [sum of first i integers, error of the sum].
+ * The "true" prefix sum is equal to the sum of the pair of numbers, but it is
+ * explicitly returned as a pair of numbers to ensure that the error portion
+ * isn't lost when subtracting prefix sums.
+ */
+function prefix_float_sum(input) {
+  const prefix_sum = [[0, 0]];
+  let sum = 0;
+  let c = 0;
+  for (let i = 0; i < input.length; i++) {
+    const cur = input[i];
+    const t = sum + cur;
+    if (Math.abs(sum) >= Math.abs(cur)) {
+      c += (sum - t) + cur;
+    } else {
+      c += (cur - t) + sum;
+    }
+    sum = t;
+    prefix_sum.push([sum, c]);
+  }
+  return prefix_sum;
+}
+
 /*
  * Probability Density Function of rates.
  * Since the PDF is continuous*, we approximate it by a discrete probability function:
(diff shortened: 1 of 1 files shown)


js/predictions.js
```diff
@@ -121,75 +193,95 @@ class PDF {
...
+    // = x(0) * 1 / 10
+
+    // Get a prefix sum / CDF of this so we can calculate sums in O(1).
+    const prefix = prefix_float_sum(this.prob);
+    const max_X = this.prob.length;
+    const max_Y = rate_decay_max - rate_decay_min;
+    const newProb = Array(this.prob.length + max_Y);
+    for (let i = 0; i < newProb.length; i++) {
+      // Note that left and right here are INCLUSIVE.
+      const left = Math.max(0, i-max_Y);
```
peter50216, line 257: nit: space around operator `-`?
(Also on the following lines)

mcpower: Done.

mcpower:
@peter50216 Does this project have a style guide (or linter, or autoformatter)? It'd be useful for future reference, although considering the size of this project it may be a bit overkill.

peter50216:
> @peter50216 Does this project have a style guide (or linter, or autoformatter)? It'd be useful for future reference, although considering the size of this project it may be a bit overkill.

I don't think there's one. It does feel a bit overkill to me too, and I just follow the style of existing codes.

theRTC204:
> @peter50216 Does this project have a style guide (or linter, or autoformatter)? It'd be useful for future reference, although considering the size of this project it may be a bit overkill.

No styleguide or linter at this point, no.

mikebryant:
And @peter50216 thanks for reviewing, because it's all going *whoosh* over my head right now :)

PR #278, opened 2020-04-26
Adding multi-island support

DevSplash:
1. It is possible to record the price of multiple islands. 
2. Add island names to distinguish them. 
3. Permalinks data can be saved into localStorage.


KKKKaiZhu:
Well done DevSplash! Very good feature!
However, if you look at my pull request #275, you will notice we are both adding a button to the right bottom corner and the buttons are going to overlap if both are merged.
My personal opinion: it would be very nice if you can move your button to the left (or somewhere else?) since its height seems not fixed, which means it could affect other buttons added to the left in the future.
If you have a better solution let me know, I am very happy to change my code (which is way shorter than yours!!!)

theRTC204:
What considerations were made for mobile users in this change? A large stack of floating buttons is not likely a great experience - the single button for dark mode is already not great.

DevSplash:
> What considerations were made for mobile users in this change? A large stack of floating buttons is not likely a great experience - the single button for dark mode is already not great.

Here is demo on mobile device.

DevSplash:
> Well done DevSplash! Very good feature!
> However, if you look at my pull request #275, you will notice we are both adding a button to the right bottom corner and the buttons are going to overlap if both are merged.
> My personal opinion: it would be very nice if you can move your button to the left (or somewhere else?) since its height seems not fixed, which means it could affect other buttons added to the right in the future.
> If you have a better solution let me know, I am very happy to change my code (which is way shorter than yours!!!)

An easy way is that you change the bottom to 85px and then I modify my button position. 
But the better way is to put them in a div tag and use relative positioning.

KKKKaiZhu:
> 
> 
> > Well done DevSplash! Very good feature!
> > However, if you look at my pull request #275, you will notice we are both adding a button to the right bottom corner and the buttons are going to overlap if both are merged.
> > My personal opinion: it would be very nice if you can move your button to the left (or somewhere else?) since its height seems not fixed, which means it could affect other buttons added to the right in the future.
> > If you have a better solution let me know, I am very happy to change my code (which is way shorter than yours!!!)
> 
> An easy way is that you change the bottom to 85px and then I modify my button position.
> But the better way is to put them in a div tag and use relative positioning.

I agree. Let's do the easy way first, I have changed it to 85px.

mikebryant:
I think this is out of scope for the project, as it says in the README. 

> Turnip Prophet is not a way to store multiple people's islands

Storing multiple islands adds complexity to the UI, but doesn't actually solve anything about multiple people all wanting to put island data in and see it compared together.

I think I've said this on Discord, but not yet in the issue tracker, one thing I am interested in is being able to graph a set of multiple islands submitted by query string. (Then I can use a spreadsheet or similar to have access control, allow multiple people to edit, and be able to click on a link to see the combined graph of all people.)

DevSplash:
> I think this is out of scope for the project, as it says in the README.
> 
> > Turnip Prophet is not a way to store multiple people's islands
> 
> Storing multiple islands adds complexity to the UI, but doesn't actually solve anything about multiple people all wanting to put island data in and see it compared together.
> 
> I think I've said this on Discord, but not yet in the issue tracker, one thing I am interested in is being able to graph a set of multiple islands submitted by query string. (Then I can use a spreadsheet or similar to have access control, allow multiple people to edit, and be able to click on a link to see the combined graph of all people.)

This feature is aimed at people with multiple devices who can record prices for multiple islands simultaneously. it would be difficult to implement without this feature.  (At least I think it's a useful feature.)
README is just a description to the project, and it is entirely possible to modify it if new features are added to it. 

In addition, the feature of displaying multiple islands on the same chart can be expanded on this feature.

theRTC204:
I think you misunderstood. Mike was being polite, but indicating that your PR is not something that he sees as a feature he wants in his project.

The README says what it says because the project maintainer (Mike) has defined guidelines for contributors on what type of things he wants in the project, and which ones he doesn't.

DevSplash:
> I think you misunderstood. Mike was being polite, but indicating that your PR is not something that he sees as a feature he wants in his project.
> 
> The README says what it says because the project maintainer (Mike) has defined guidelines for contributors on what type of things he wants in the project, and which ones he doesn't.

I see. This PR can be turned off if the project doesn‘t need this feature. 
(But I still think this feature is useful. 😄)

mikebryant:
The use-case of a single person with multiple Switches is not something I'd actually considered. I'm going to think on this..

theRTC204:
> The use-case of a single person with multiple Switches is not something I'd actually considered. I'm going to think on this..

Is that really a likely usecase? Sure, some families might have multiple Switch consoles, with multiple copies of the game, but those families probably also at least have multiple devices they could use to track prices on this app. Maybe not multiple home computers, but at least a computer, phone or table.

Seems like a super niche edge case for such a large change to the project.

jonathan-irvin:
+1 For this. My wife and stepson each have switches and it's usually easier to keep track on one device than all of us plugging in the numbers.

theRTC204:
Yes, and that's why support for query strings was added. The small number of people who have multiple Switch's in their home, and a single device for tracking, can enter their numbers into something like a spreadsheet, and automatically generate links to TP for each island.

```
=HYPERLINK(CONCATENATE("https://turnipprophet.io?prices=", ENCODEURL(JOIN(".", B5:N5)), IF (ISBLANK(O4), "", CONCATENATE("&pattern=", LOWER(SUBSTITUTE(O4, " ", "-"))))), "Turnip Prophet")
```

There is an example of a working formula which will do exactly that.

Adding significant complexity to the website to serve the needs of a very small group of people is not something that we're interested in doing right now.

Mike has mentioned he's thinking about ways to enhance the query string functionality to allow data from multiple islands to be fed in that way, but that approach would not store it - it would be for display only.

PR #284, opened 2020-04-26
fix: Decrease precision for tiny numbers

### LINKED ISSUE #216: What a surprise! I see "e" in the site
Excuse me, I'm not good at English.

Look at this link. [https://turnipprophet.io/?prices=107.96.91.86.81........&pattern=3](https://turnipprophet.io/?prices=107.96.91.86.81........&pattern=3)

"Fluctuating" is "2.39e-7%" ! as 0% ?

mikebryant:
When the numbers are so small then end up in e notation is pushes
everything too far to the right. Give up when it's too small to be
really useful.

Fixes #216

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -259,6 +259,21 @@ const getPriceClass = function(buy_price, max) {
   return "";
 }
 
+const displayPercentage = function(fraction) {
+  if (Number.isFinite(fraction)) {
+    let percent = fraction * 100;
+    if (percent >= 1) {
+      return percent.toPrecision(3) + '%';
+    } else if (percent >= 0.01) {
+      return percent.toFixed(2) + '%';
+    } else {
+      return '<0.01%';
+    }
+  } else {
+    return '—'
+  }
+}
+
 const calculateOutput = function (data, first_buy, previous_pattern) {
   if (isEmpty(data)) {
     $("#output").html("");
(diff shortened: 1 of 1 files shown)


js/scripts.js
```diff
@@ -259,6 +259,21 @@ const getPriceClass = function(buy_price, max) {
   return "";
 }
 
+const displayPercentage = function(fraction) {
+  if (Number.isFinite(fraction)) {
+    let percent = fraction * 100;
+    if (percent >= 1) {
+      return percent.toPrecision(3) + '%';
+    } else if (percent > 0.01) {
```
peter50216, line 267: Should this be 0.1?

theRTC204: 0.01 seems right to me. We want anything less than. 1, but greater than 0.01 to appear as itt's actual percentage value.

That being said, now that I look at it, we probably want `>=` instead of `>`.

peter50216: Then should the "else" part be "<0.01%"? Or it might be weird that both "0.02%" and "<0.1%" can appear.

theRTC204: Oh I didn't even catch that. I assume the else was meant to be `<0.01%`

mikebryant: Good catch, thanks. Fixed :)

Issue #288, opened 2020-04-27
First-Time Buyer Buy Price


zackman0010:
When first-time buyer is enabled, the Sunday box is disabled. However, I've noticed that the results shown on the table still show all the different possibilities for the Sunday buy price. I've also noticed that the boxes do not get their colored highlighting without a value entered in the Sunday box.
Even if the buy price is not used in the calculation on first-time buyer (my assumption, no idea if that's why it's disabled), would it be possible to still allow a value to be entered to further filter the results of the table and to enable the colored boxes?

mikebryant:
> I've noticed that the results shown on the table still show all the different possibilities for the Sunday buy price.

This is because the buy price you had gets regenerated on Monday morning, and it uses this hidden buy price for the week's calulations

> Even if the buy price is not used in the calculation on first-time buyer (my assumption, no idea if that's why it's disabled), would it be possible to still allow a value to be entered to further filter the results of the table

No point in having a value, it can't filter the table, you don't know the hidden price

> I've also noticed that the boxes do not get their colored highlighting without a value entered in the Sunday box.

That's odd - we should fix that, it should be coloured regardless

mikebryant:
Oh, I see, we're actually basing that off the buy price. Hmmmm.

Maybe we do need to let someone enter the buy price anyway then.. *ponders*

Might be worth renaming the buy price column too then, or hiding it, in the first-time case

zackman0010:
![image](https://user-images.githubusercontent.com/9540563/80389430-b0ac0680-8870-11ea-9371-06daeb88316e.png)
So the Sunday value listed here in this table doesn't actually affect the rest of the prices?
If that's true, then could the Sunday value be hidden entirely for first-time (Which I see you actually commented as I was typing this, lol)?

mikebryant:
The Sunday value there is what the rest of the prices are actually computed from - it's just not related to the price you actually saw

zackman0010:
Ah, that makes sense. Thank you!

gallexi:
> No point in having a value, it can't filter the table, you don't know the hidden price

Oh, that makes sense! I too was confused by this, and would +1 either a little message explaining this or just hiding the buy-price column for first-time buyers (or both!) =]

Thanks so much for this app!

chrisgchambers:
[...]

PR #293, opened 2020-04-27
Table Styling for First-Time Buyers

chrisgchambers:
Table styling was broken because buy_price is not entered by first-time buyers. I allowed the Sunday minimum value from the predictions to substitute in this case for styling purposes.

Fixes the bug portion of issue #288


chrisgchambers:
Tweaking to style table whenever buy price is empty for both first-timers and regular users

PR #296, opened 2020-04-27
fix: Allow people to input the buy price even as a first time buyer

### LINKED ISSUE #288: First-Time Buyer Buy Price

mikebryant:
This will give us a basis for colouring. Also remove the Sunday prices from the table, to avoid confusion and questions when people ask why the Sunday price is different to the one they put in. This has the added bonus of making the table smaller, and doesn't really remove any relevant information for the future.

Fixes #288


locales/pt-BR.json
```diff
@@ -25,7 +25,7 @@
     "small-spike": "Pequeno Pico"
   },
   "prices": {
-    "description": "Qual foi o preço dos nabos esta semana em sua ilha? <i>(Se esta é a primeira vez que compra nabos, este campo ficará desativado)</i>",
```
theRTC204, line 28: Nitpick: you yanked the `?` out of this line.

mikebryant: Bah, thanks for spotting that!

Fixed

PR #300, opened 2020-04-28
Fix incorrect "Guaranteed Minimum"

RJacksonm1:
Spotted something odd where my "Guaranteed minimum" was greater than my current prices

https://turnipprophet.io?prices=96.87.83.73.74........

Digging through the code, I think this line is the mistake. The guaranteed minimum should be the lowest of day-by-day minimums, right?

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -954,7 +954,7 @@ class Predictor {
         weekMins.push(poss.prices[poss.prices.length -1].min);
         weekMaxes.push(poss.prices[poss.prices.length -1].max);
       }
-      poss.weekGuaranteedMinimum = Math.max(...weekMins);
+      poss.weekGuaranteedMinimum = Math.min(...weekMins);
       poss.weekMax = Math.max(...weekMaxes);
     }
 


peter50216:
Nope, the guaranteed minimum is the minimum guaranteed price. That is, in the most unlucky situation, the minimum price you would ever got if you sell your turnips at the best price you can see.

So it's correct that you would at least be able to sell your turnips for at least 87 each at some point in this week.

RJacksonm1:
I think I get ya, but boy does it make my brain hurt.

Looking at the week ahead, I will definitely get at least one price at or above 87. Is that what we're saying?

But it's also entirely possible that I can miss that window, and end up with a lower minimum? E.g. looking at the link above, if I wait until Saturday I could end up forced into two below-guaranteed-minimum prices.

I think that's where I'm getting confused with the "Guaranteed" terminology – the fact I can still end up with a price below that.

theRTC204:
Your understanding seems correct, yes; and I agree we have struggled with the naming of that column for a while now. Some refer to it as "least best" or "future minimum".

You are correct is saying if you miss a price period entirely, you could end up with a worse value than what is shown. That summary column is not meant to tell you _when_ to sell; that's what the rest of the table is for.

RJacksonm1:
Alright, thanks for the explanations all!

PR #302, opened 2020-04-28
Ensure attribution of all contributors

MasterKale:
The current method of pulling contributors (implemented in #84) only displays 30 contributors because the Github API return value caps out at 30 individuals. As a result, a third of all contributors to date are not given proper attribution. I believe the popularity of the project has made it such that there's no choice but to start maintaining a list of contributors within the repo.

This PR establishes a list of contributors in a dedicated JSON file within the repo. The API request has been refactored to pull and display this list at runtime instead of the one returned by the Github API.

And while the current list of 44 contributors is included in this PR, future contributors will need to add themselves alphabetically as part of their first PR. These instructions are included in a new README section to help ensure that future contributors are properly attributed.


README.md
```diff
@@ -58,6 +58,13 @@ Please make sure **not to translate** "Turnip Prophet" and include the new langu
 
 If you have any remaining questions, feel free to stop by the Discord server and ask. 
 
+## Contributing
+
+If you submit a PR and your account isn't listed in **contributors.json**, please add it in (alphabetically) as part of your PR! Your entry should look like this:
+
+```json
+["UsernameHere", "https://github.com/UsernameHere"]
```
mikebryant, line 66: nit: Should probably end with a comma, which will be the most common case

README.md
```diff
@@ -58,6 +58,13 @@ Please make sure **not to translate** "Turnip Prophet" and include the new langu
 
 If you have any remaining questions, feel free to stop by the Discord server and ask. 
 
+## Contributing
+
+If you submit a PR and your account isn't listed in **contributors.json**, please add it in (alphabetically) as part of your PR! Your entry should look like this:
```
mikebryant, line 63: Can you make `contributors.json` a link to the file please?

MasterKale: This now links to the file as requested 😸

README.md
MasterKale: Done and done :)

MasterKale:
It's weird to me that end user devices are forced to perform N number of additional API requests _on every page load_ for the sake of developer convenience. That's a lot of network traffic to queue up every time the user accesses the site.

This PR would convert these N requests to Github's API into 1 request for a single JSON file, saving end user battery life at the expense of a contributing developer's additional minute or two of time to include an entry for theirself in the JSON file. I'd argue this is worth it in the name of better UX.

theRTC204:
We're not talking about dozens of API calls to a data heavy endpoint. We're talking about 2, and perhaps one day 3 calls to an endpoint that returns 4KB of data.

By comparison, visiting Google's search homepage has a larger total data size, and makes more raw network requests.

Turnip Prophet - 27 requests - 1.3 MB of data
Google - 32 requests - 1.9 MB of data

I'm not saying your points aren't correct, but they are definitely blown out of proportion here. Be that as it may, ultimately @mikebryant will decide whether he agrees with my comments or not. 😛

MasterKale:
I'll be honest, when I came across #84 I shook my head a bit. Adding users directly to the HTML document was the most simple solution, requiring the least amount of maintenace. Now there's (to @theRTC204's point an admittedly small number of) additional Ajax requests being made, JavaScript to parse and add in new DOM elements, and now logic needed to parse and recurse on the `Link` header value...it just means more complexity and code in this repo that has to be maintained.

If you want to stick with hitting the Github API in the name of not asking devs to remember to manually add themselves to a contributors list, then what I honestly think should happen here is a build process gets introduced that pulls in the users from that API and embeds them into the HTML.

This too seems like too much, though, when the alternative is the simple addition of a one-line HTML diff to a PR when someone contributes for the first time 🤔

akccakcctw:
> Implementing the All Contributors spec by hand is a laborious process.

We don't have to do all the jobs because there is a ready-made GitHub bot here.

https://allcontributors.org/docs/en/bot/overview


I never try it before but it looks to fit our needing.

DevSplash:
It takes more time to manually maintain the list. I agree with @theRTC204.

Tekaoh:
Not that my opinion counts for much, but I don't personally like this solution either. Using the API is so much nicer. Granted, it's not a lot of effort to add your name to the file, but most probably won't even realize that's something they should do. When I submitted my first PR I didn't even think about whether my name would be included on the website or not, but I was delighted when I discovered later that it automatically was.

mikebryant:
Thank you for submitting this, but I'm going with #311

PR #305, opened 2020-04-29
adding delay after user input to prevent jank as chart refreshes

Jollus88:
I've noticed that there's a bit of lag whenever typing in the input fields, and thought I could help. It's not always so bad on desktop, but especially janky on mobile.

This timeout delay stops the chart from rendering until 1000ms after the last keystroke, and helps ensure that sets of data can be input without re-rendering the chart after each keystroke.

### DIFF
--- a/js/translations.js
+++ b/js/translations.js
@@ -51,6 +51,15 @@ i18next
   });
   // init set content
   $(document).ready(initialize);
-  $(document).on('input', updateContent);
+
+  let delayTimer;
+  $(document).on('input', function() {
+    // adding short delay after input to help mitigate potential lag after keystrokes
+    clearTimeout(delayTimer);
+    delayTimer = setTimeout(function() {
+      updateContent();
+    }, 1000);
+  });
+
   $('input[type = radio]').on('change', updateContent);
 });
--- a/js/translations.js
+++ b/js/translations.js
@@ -58,7 +58,7 @@ i18next
     clearTimeout(delayTimer);
     delayTimer = setTimeout(function() {
       updateContent();
-    }, 1000);
+    }, 500);
   });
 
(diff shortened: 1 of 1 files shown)


js/translations.js
```diff
@@ -51,6 +51,15 @@ i18next
...
   $(document).ready(initialize);
-  $(document).on('input', updateContent);
+
+  let delayTimer;
+  $(document).on('input', function() {
+    // adding short delay after input to help mitigate potential lag after keystrokes
+    clearTimeout(delayTimer);
+    delayTimer = setTimeout(function() {
+      updateContent();
+    }, 1000);
```
theRTC204, line 61: 1 second debounce is a bit execessive I think. 500ms should be more than enough.

Jollus88: Sure, I've just tested with 500ms and that should be OK too. Have updated the PR.

DevSplash:
Would it be better to use change licenser?
`$('input:not([type=text])').on('change', () => update());`
It might also need to modify here. 
https://github.com/mikebryant/ac-nh-turnip-prices/blob/dd91bf1ea4cc75c0f29f20793370b3b16135b3b2/js/scripts.js#L84

theRTC204:
The difference between the two is `input` is only fired when a change is made via the interface, while `change` is fired any time the contents change.

In our use case, since we're not programmatically changing the value of the inputs, there's no need to change the listening. We've been using `input` the entire time - no need to fix what isn't broken.

mikebryant:
Wait. This PR was to the wrong branch. hmm

Jollus88:
Yes, the merge should have gone to the WIP branch rather than straight to Master. Sorry if that was incorrect.

Issue #314, opened 2020-05-01
Predictions are calculated twice when radios are toggled


TarzanInAJar:
**Things to check**
- Have you forced a refresh in your browser (e.g. ctrl+F5 in Chrome/Firefox)?
  - yes
- Have you made sure the first time buyer option is correct? If you're unsure, try it both ways
  - yes
- Have you time-travelled this week? Travelling backwards resets the prices, and therefore you must not put prices in from both before & after the time-travel
  - no

**Describe the bug**
It appears that changing a radio toggle (First-Time Buyer, Previous Pattern) triggers `updateContent()` twice - once on the `input` event, another on the `change` event.

**To Reproduce**
1. Set a breakpoint on `updateContent()`
1a. Alternatively, have both the radio buttons and the chart in view (without having to scroll)
2. Change either First-Time Buyer or Previous Pattern options
3. If using the debugger, note that the `updateContent()` function is fired twice
3a. Alternatively, if viewing the chart, notice that the chart is rendered twice in quick sequence

**Permalink to your prices**
https://turnipprophet.io?prices=.50.100.150.200........

**Expected behavior**
updateContent() should only be fired once.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 19.10
 - Browser: Firefox
 - Version: 75.0 64-bit

**Additional context**
It doesn't appear that listening to the `change` event in addition to the `input` event is necessary - radios all seem to trigger just fine on just `input`. I think this can be fixed by simply removing [this line](https://github.com/mikebryant/ac-nh-turnip-prices/blob/f8a223ad533c02267c7ada23062b80a33acffe37/js/translations.js#L66).

theRTC204:
Thanks for the report. Seems the recently added denounce is what revealed the visible double rendering, because one of them is delayed 500ms.

PR #315, opened 2020-05-01
#314 Fix Predictions are calculated twice when radios are toggled

daniel-giralt-len:
Fixes https://github.com/mikebryant/ac-nh-turnip-prices/issues/314

Since radio buttons are subject to both the `document.input` event and the `input[type = radio].change` event the content is updated twice every time a radio group's value is changed.

I did not remove the change event altogether; although it should work , [MDN ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) says: 
> For <input> elements with type=checkbox or type=radio, the input event should fire whenever a user toggles the control, per the HTML5 specification. However, historically this has not always been the case. Check compatibility, or use the change event instead for elements of these types.

So I went for a guard clause on the overall input event.

A further suggestion of mine is to bind the event to every input tag and not the document, but I haven't analyzed the project enough to make that PR.

Anyway, cheers!

### DIFF
--- a/js/translations.js
+++ b/js/translations.js
@@ -55,7 +55,9 @@ i18next
   $(document).ready(initialize);
 
   let delayTimer;
-  $(document).on('input', function() {
+  $(document).on('input', function(event) {
+    //prevent radio input from updating content twice per input change
+    if(event.target.type === 'radio'){ return }
     // adding short delay after input to help mitigate potential lag after keystrokes
     clearTimeout(delayTimer);
     delayTimer = setTimeout(function() {


theRTC204:
In terms of binding the listener to the document vs the elements themselves, since ever input on the page ultimately fires the same update function, binding to the window is probably more efficient being a single listener.

If the page was more complex, and contained inputs that weren't meant to fire this same function, then a more specific listener binding would be ideal.

daniel-giralt-len:
^ Sorry about these, some branch shenanigans happened

PR #318, opened 2020-05-01
Fix #317 -- Margin discrepancy light dark mode

### LINKED ISSUE #317: (Minor)(UI) Margin discrepancy between light and dark mode
**Things to check**
  - yes
  - yes
  - yes

**Describe the bug**
Light mode has 8px of margin while darkmode has none (as it is position:absolute). This makes switching between modes a tad jarring as the whole layout moves slightly.

**To Reproduce**
1. Open the page
2. Switch modes repeatedly

**Expected behavior**
Same amount of spacing and structure between modes

**Screenshots**

 - OS: W7
 - Browser Chrome
 - Version 81

Jollus174:
https://github.com/mikebryant/ac-nh-turnip-prices/issues/317

Making the darkmode-background `position: fixed;` and removing the `position: absolute;` and `margin: 0;` from the body when darkmode is activated is a clean fix for this. There's also some redundant body styles that I removed.

As a side note, I did notice this 'waves' SVG is interfering with darkmode. Not sure if it should still be there?
image

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -568,23 +568,12 @@ input[type=number] {
   outline: none;
 }
 
-div.darkmode-background{
+div.darkmode-background {
   background: #fef0e3;
   background-image:
     radial-gradient(#fff 20%, transparent 0),
     radial-gradient(#fff 20%, transparent 0);
-  position: inherit;
-}
-
-body.darkmode--activated{
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  justify-content: center;
-  font-family: 'Varela Round', sans-serif;
-  width: 100%;
-  margin: 0 0 0 0;
-  position: absolute;
+  position: fixed;
 }
 
 body.darkmode--activated div[class^="dialog-box"],
--- a/css/styles.css
+++ b/css/styles.css
@@ -573,6 +573,10 @@ div.darkmode-background {
   background-image:
     radial-gradient(#fff 20%, transparent 0),
     radial-gradient(#fff 20%, transparent 0);
+  position: static;
+}
+
+.darkmode--activated .darkmode-background {
   position: fixed;
 }
 


theRTC204:
While this does address the position shift, it seems to introduce a weird artifact behind the page elements.

image
image
image

As you scroll down the page there is a little black oval that stays fixed to the center of the page. I suspect that this is the dotted effect that exists on the light mode which is being thrown out of wack by the fixed positon.

css/styles.css
```diff
@@ -568,23 +568,12 @@ input[type=number] {
...
-}
-
-body.darkmode--activated{
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  justify-content: center;
-  font-family: 'Varela Round', sans-serif;
-  width: 100%;
-  margin: 0 0 0 0;
```
theRTC204, line 586: Seems as though simply changing the margin value to include the 8px that exists in the light mode fixes the shift, without introducing the weird oval artifact mentioned in my earlier comment.

```css
margin: 8px 0 0 0;
```

theRTC204: Hmm, nope I was wrong. That causes an ever so slight white edge at the top, and doesn't correct the shift when scrolled to the bottom of the page.

However... leaving the margin at 0, and instead adding a bit of padding does work.

```css
body.darkmode--activated {
  display: flex;
  ...
  margin: 0 0 0 0;
  padding: 8px 0;
}
```

Jollus174: Not quite, adding padding to the top + bottoms doesn't stop the gutters from disappearing at the sides of the page when darkmode is activated.
image

I think this problem is caused switch to `position: absolute;` ignoring the existing gutters. I'm just updating the branch now, if you do a pull and take a look, that background should be back now (so THAT'S what the `position: inherit;` was for)

theRTC204:
I think we should hold on this until further discussion around #325 occurrs. This change wouldn't be necessary if we end up removing darkmodejs entirely.

Jollus174:
Sure, that's fine. I'll withdraw this PR.

Issue #233, opened 2020-04-22
Dark mode graph


Mienshao222:
I appreciate the new dark mode, but the graph line us almost invisible. When you get a chance, could you please fix the colour? Thankyou

theRTC204:
If I'm being honest, there's a lot a out the dark mode that could use some love.

binarykittenDE:
Yeh, the dark mode is a nice touch, but the colors need some fix up!

Issue #254, opened 2020-04-24
Alternate color scheme for colorblind users


boosterbeetle:
Just noticed the new colors to represent good/bad prices but as someone with red/green colorblindness it's pretty difficult to distinguish between the shades of green. Also checked out the dark mode equivalent and the shades of purple? are also difficult to tell. Not a deal breaker by any means but just leaving some feedback in case you want to add support for other schemes in the future. All in all, great work.

theRTC204:
Great suggestion. There are a handful of things we should do for accessibility.

davidcorry:
I've made an initial pass at a higher contrast theme, that should (hopefully) work across the different forms of colorblindness. I don't have any colorblindness, though, so I was relying on an app to simulate it for me. Any feedback from folks who this would benefit would be great! (@boosterbeetle)

I haven't made a dark mode, but depending on interest I can see about how to handle the logic for that.

<img width="849" alt="image" src="https://user-images.githubusercontent.com/265951/82398317-7664fd80-9a07-11ea-94cd-f73dc2cfb993.png">
<img width="1235" alt="image" src="https://user-images.githubusercontent.com/265951/82398351-88df3700-9a07-11ea-8583-07f52b2a888e.png">

boosterbeetle:
@davidcorry Screenshot looks good! Easier to tell between different sets of values now.

davidcorry:
Rad! If anything needs adjusting, let me know. I’ll likely come up with a dark mode version in the next day or two and hopefully we can get it in soon.

boosterbeetle:
Yeah, not sure if others want to pitch in but for me, I had a hard time distinguishing between the best and second best values in the previous color scheme (shades of green used were very similar). It's solved here though as one of the blue is much darker than the other one. Don't know if that still causes problems with other users though.

PR #325, opened 2020-05-04
CSS-based dark mode

### LINKED ISSUE #70: Dark Mode

### LINKED ISSUE #317: (Minor)(UI) Margin discrepancy between light and dark mode

davidcorry:
This creates a new CSS-based dark mode that automatically responds to the device's preferred mode. This fixes #317, #233, and should close #70.

I came up with the color scheme myself, trying to make it an appropriate match to the light theme. I'm totally open to changes. One advantage of this method is that it's really easy to update colors.

Since there's no more need for darkmode.js, the one downside is that there's not a toggle button for folks who want to override which color scheme gets used, but it's entirely possible to set up a button again. If there's interest in that, I can figure something out!

The one thing I want to note is that there's a lot of different shadow color values - I thought about simplifying these, but ended up just leaving them in for now.

Here's some pics of what it looks like:

image
image

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -64,58 +64,56 @@
   --table-range4: hsl(0, 80%, 85%);
 }
 
-@media (prefers-color-scheme: dark) {
-  :root {
-    --bg-color: #1A1A1A;
-    --bg-dot-color: #222;
+[data-theme="dark"] {
+  --bg-color: #1A1A1A;
+  --bg-dot-color: #222;
 
-    --shadow-3: rgba(255, 255, 255, 0.03);
-    --shadow-15: rgba(255, 255, 255, 0.03);
+  --shadow-3: rgba(255, 255, 255, 0.03);
+  --shadow-15: rgba(255, 255, 255, 0.03);
 
-    --center-bg-color: #101010;
+  --center-bg-color: #101010;
 
-    --wave-1: rgba(16, 16, 16, 0);
-    --wave-2: rgba(16, 16, 16, 0.2);
-    --wave-3: rgba(16, 16, 16, 0.4);
-    --wave-4: rgba(16, 16, 16, 0.6);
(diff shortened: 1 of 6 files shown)


theRTC204:
I haven't done a full review yet, but I am definitely in favor of swapping darkmode.js out for a pure CSS dark mode.

That being said, `prefers-color-scheme` is still a very new media query, and while widely supported in the latest versions, it only goes back about 8-9 months for some browsers.

https://caniuse.com/#search=prefers-color-scheme

Before this is merged I would like to see a few enhancements.

- A button to toggle for personal preference and/or users on devices that don't have a system level dark mode (upon which this media query is based)
- Maintain the user preference in Local Storage, which is the default value, that falls back to the device preference if LS has no value

davidcorry:
Alright: I added a drop down menu beneath the menu selector that lets users specify their theme. I also set it up in a way that it's easy to add additional themes, if we were to come up with more accessible options. (See #254)

I don't have a device / browser currently that _doesn't_ support dark mode, but I'm fairly confident that it handles that situation properly.

davidcorry:
I found an old copy of Firefox on my computer, so I was able to test browsers that don't support automatic switching - all good. Just had to make a quick update since it wasn't loading the strings.

PR #346, opened 2020-05-06
Add expected value info to chart and table

nejstastnejsistene:
The expected value is the weighted average of each column of regular data
in the table. My island was roughly 50/50 large spike vs decreasing this morning, and without an expected value it's difficult to tell whether to cut my losses or wait  in hopes that it's the large spike.

Query params: `?prices=100.86.81.77.74.71.......&pattern=0`
Screen Shot 2020-05-06 at 13 58 42

To settle the suspense, today's PM price confirms I'm most likely getting the large spike!

Query params: `?prices=100.86.81.77.74.71.119......&pattern=0`
Screen Shot 2020-05-06 at 13 59 09

### DIFF
--- a/js/chart.js
+++ b/js/chart.js
@@ -28,10 +28,14 @@ function update_chart(input_data, possibilities) {
       data: possibilities[0].prices.slice(1).map(day => day.min),
       fill: false,
     }, {
+      label: i18next.t("output.chart.expected-value"),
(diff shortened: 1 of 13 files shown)


nejstastnejsistene:
Just started reading through some of the issues and I realized that the math here is incorrect because the the distribution between min/max is not necessarily uniform. Keeping this draft around in case it becomes useful...

PR #349, opened 2020-05-07
Adding error message for infinity turnips

ebroom:
This adds an error message for if a user gets infinity turnips. I assume that it's an input error and if not, ask users to submit a bug report. Hopefully this is a better user experience.

This PR also removes the chart if a user clears out their data.

I'm not sure exactly what to do with locale since I'm adding a string, butI'm only fluent in English. 

turnip-error-message


theRTC204 (approved):
Thank you for this awesome contribution! It was on my list of things to get to eventually.

As far as the translations go, if there is no translation the application will gracefully fall back to English. We rely on the community to update localizations.

theRTC204 (approved):

theRTC204:
Implements #386

PR #353, opened 2020-05-07
Decoupling Predictor class from i18next.

### LINKED ISSUE #201: Remove i18next from predictions.js
It would be ideal if predictions.js was once again purely the math behind all of the turnip calculations and not FoS code with libraries to import. When decoupled, the calculations can be used for other branching scripts when this project is abandoned - otherwise the calculations are tied to this library.

CatStarwind:
Cleaning up Predictor class by removing dependency on i18next. Moved translation logic to display function (``calculateOutput``) in script.js.

Closes #201

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -725,8 +725,8 @@ class Predictor {
     }
 
     // Now each day is independent of next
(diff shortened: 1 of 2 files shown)


theRTC204:
Generally speaking I'm fine with this change, but man all those semicolon's made finding the relevant changes a total pain.

CatStarwind:
f1a6680 should have the main changes by itself.

Issue #357, opened 2020-05-10
All the contributors


Tekaoh:
I just noticed that there are only 30 contributors listed on https://turnipprophet.io/ while the GitHub repository lists 53 contributors now. The is because the API call currently used only returns page 1 of the contributors, and you get 30 contributors per page.

https://github.com/mikebryant/ac-nh-turnip-prices/blob/a400ee92180459f07052022133c4d0fcfd3209d7/js/contributors.js#L4

To get the rest of the contributors, you would need `https://api.github.com/repos/mikebryant/ac-nh-turnip-prices/contributors?page=2`, and once there are over 60 contributors, page 3 and so on.

It seems that [v3](https://developer.github.com/v3/repos/collaborators/#list-collaborators) of GitHub's API can respond with URLs for the next page and the last page, so it would be possible to build logic to loop through all available pages and display the entire list of contributors.

Is this something we want to do? Or is it preferable to only list the first 30 contributors?

John-Kryspin:
I think this is a great idea! I added a PR as a starting point. I think right now this PR solves the issue, but perhaps if even more people contribute we might want to look into adding a show more option in the future, if the amount of contributors scales very high.

theRTC204:
We now have 3 active PRs to address this concern... #302, #311 and now #358.

Tekaoh:
Oh, I'm sorry... I shouldn't have bothered opening an issue if that's the case. I scanned through the issues, but didn't think to look at open PR's. My bad

theRTC204:
I appologize, I didn't mean to suggest raising the issue was a bad idea. It's good to have the issue open for tracking until one of the open PRs is accepted!

Tekaoh:
Sounds good. Thanks for linking the existing PRs!

mikebryant:
Thanks for raising this, this is now fixed :)

PR #358, opened 2020-05-10
use github's response header "link" to loop through contributors API

John-Kryspin:
References #357 This recursively appends the contributors. 
This resolves only the top 30 contributors being shown.


mikebryant:
Thank you for submitting this, but I'm going with #311

PR #359, opened 2020-05-11
Change the Japanese translation of "chance"

patilise:
"チャンス" is not totally wrong, but "確率" (direct translation of "probability") sounds more natural here.


patilise:
Withdraw and resubmit

PR #369, opened 2020-05-14
Fixed off-by-one in calculation of small spike.

peter50216:
The peak is always at least one higher then the previous value in small spike.


rda2:
Shouldn't the low price be increased by 1 as well?

PR #374, opened 2020-05-17
Change lang attribute according to chosen language

patilise:
This is a change such that the lang attribute of html tag does not remain "en" when a non-English language is chosen.


theRTC204 (approved):
Thank you for the contribution!

Issue #375, opened 2020-05-18
App isn't saving anything 


baseblade:
**Things to check**
- Have you forced a refresh in your browser (e.g. ctrl+F5 in Chrome/Firefox)?
  - using mobile app
- Have you made sure the first time buyer option is correct? If you're unsure, try it both ways
  - yes
- Have you time-travelled this week? Travelling backwards resets the prices, and therefore you must not put prices in from both before & after the time-travel
  - /no

**Describe the bug**
yesterday the app completely cleared everything you said of me having to delete all reset and now there is nothing saving when I'm putting the prices and all my statistics have gone

**To Reproduce**

**Permalink to your prices**
Please click the `Copy Permalink` button and paste it here

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 - Device: Google Pixel 3XL
 - OS: unsure
 - Browser chrome 
 - Version ?

**Additional context**

Tekaoh:
This is happening to me too using Chrome on Windows 10. I lost my save data late last week but then it appeared to be saving once again. But I lost my save data again this morning. This has never happened before. Something must have changed

murielg:
Happening to me as well, onMac 10.14.5 Chrome v81, and iOS 13.3 Safari .

Hattoff:
I think this may have something to do with this statement in your scripts.js file in the update function:
```
  if (!window.populated_from_query) {
    updateLocalStorage(prices, first_buy, previous_pattern);
  }
```
If someone used a URL to generate a portion of their data, then any new input will never be saved. A simple refresh of the page doesn't reveal this problem, as the browser will cache data input into these fields. This becomes more evident, however, on mobile devices, as they don't seem to keep pages persistent, and often force-refresh the browser when users close or leave the application for an extended period of time. I believe this is why some users report this problem intermittently, as though it happens randomly.

To replicate this behavior on PC:

- Open this copied permalink:

[https://turnipprophet.io?prices=109.139.123.321.........&pattern=3](url)

- Edit some values.
- Hit F5 and note that the values stay in place.
- Now hit CTRL + F5 and notice that the query was re-loaded and previously entered data was never actually saved.

I believe by updating that if statement on the update function in the scripts.js file to something like the following:
```
  if (!window.populated_from_query) {
    updateLocalStorage(prices, first_buy, previous_pattern);
  } else {
[...]

PR #377, opened 2020-05-18
Modified initialize function; data saving bug.

Hattoff:
Updated the initialize function to check for URL parameters; if it finds any, it caches the data to local storage and then reloads the page without the query information in the URL.

Removed getPrevious function. Updated code comments.

Addresses bug [#375](https://github.com/mikebryant/ac-nh-turnip-prices/issues/375).

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -68,7 +68,15 @@ const fillFields = function (prices, first_buy, previous_pattern) {
 
 const initialize = function () {
   try {
-    const previous = getPrevious();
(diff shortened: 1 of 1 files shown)


Tekaoh:
This might not actually fix #375 
See my comment on that issue: https://github.com/mikebryant/ac-nh-turnip-prices/issues/375#issuecomment-631476484

Hattoff:
Closing. Looks like this might not address the underlying problem.

Issue #381, opened 2020-05-19
Data not saved between sessions.


Skylortrexler:
**Things to check**
- Have you forced a refresh in your browser (e.g. ctrl+F5 in Chrome/Firefox)?
  - yes
- Have you made sure the first time buyer option is correct? If you're unsure, try it both ways
  - yes
- Have you time-travelled this week? Travelling backwards resets the prices, and therefore you must not put prices in from both before & after the time-travel
  - no

**Describe the bug**
The entered user data is not being saved. 

**To Reproduce**
1. Enter data in any field
2. Close browser
3. Reopen website
4. Data is gone

**Permalink to your prices**
https://turnipprophet.io?prices=107...105.........&pattern=3 not like it matters because it will be deleted when I reopen the browser

**Expected behavior**
I expected that my data will be saved like it has been for the last few weeks. Hardly helps have a tracker if I have to manually track it as well.

**Desktop (please complete the following information):**
 - OS: Win10 x64
 - Browser: Chrome 
 - Version:  81.0.4044.138 (Official Build) (64-bit)

**Additional context**
Has been deleting my data every time since 5/17

C00kiekiller:
My wife and I have exactly the same problem.

3nyder:
Same problem here, Chrome on Android, on two different devices, phone and tablet same issue 

It was saving before (from Sunday until Tuesday morning) but we lost all info today

First time we encountered this issue, past weeks everything was ok

Mr-dotcom145:
I agree, this started happening 5/17 on my account as well.  Used to work wonderfully.

GldRush98:
Same problem, started Sunday as well. Used it for weeks and it always remembered. I have lost the data I've entered every single day this week. I have had to start writing my numbers down and putting them back in to the app every morning.

acidaris:
I suspect this is an issue from #325, as when looking at the data stored in localstorage on my desktop computer when my data reset, I had a `darkmode = true`  value which is a sign of https://github.com/sandoche/Darkmode.js (the old dark mode) instead of the `theme` value which is the new one.

Possibly if a user has the "old" format of data during initialization of the webpage, something breaks, and data gets cleared?  I'm attempting to reproduce.

Tekaoh:
Duplicate of #375

NtheOtaku:
It's been a problem for me too. What I find strange though is that my browser has cookies saved for the site.
[...]

PR #385, opened 2020-05-20
Fix data loss by preventing a call to update from running before data has been initialized.

acidaris:
Debugging through this failure, i found occasionally the "update" call would be triggered by 
[Language Change events](https://github.com/mikebryant/ac-nh-turnip-prices/blob/master/js/translations.js#L52-L54) before the old data had been initialized.

this would cause the previously loaded data to be replaced with an array of nulls.

This should address #375 / #381

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -36,6 +36,10 @@ const checkRadioByValue = function (radio_array, value) {
   radio_array.find(radio => radio.value == value).checked = true;
 };
 
+const state = {
+  initialized: false,
+};
+
 const sell_inputs = getSellFields();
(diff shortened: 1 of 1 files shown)


Tekaoh:
@mikebryant Just want to call your attention to this one. If you have a moment to take a look at this, it would be great if we can fix this critical bug.

Tekaoh:
I'm running a copy of mikebryant/master with this merged in on my own server, and it seems to have entirely fixed the problem. +∞ to merge this. @mikebryant @theRTC204

Tekaoh:
Closes #375 and Closes #381

Issue #386, opened 2020-05-20
Let user know their inputs are invalid


John-Kryspin:
**Is your feature request related to a problem? Please describe.**
Sometimes when entering data you might enter it improperly, which causes the predictor to fail.

**Describe the solution you'd like**
A notification should be shown to the user if they enter data that the predictor cannot find

**Describe alternatives you've considered**

**Additional context**
We can use a simple alert to let the user know that their inputs are incorrect.

theRTC204:
This was implemented in #349

PR #387, opened 2020-05-20
add alert message upon invalid inputs

John-Kryspin:
Solves #386
image

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -926,6 +926,10 @@ class Predictor {
         break;
       }
     }
+    if(generated_possibilities.length ===0){
+      //No pattern could be found for the prices given
+      window.alert( i18next.t("textbox.invalid-prices"))
+    }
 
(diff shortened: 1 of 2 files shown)


Lego6245:
Not the hugest fan of `alert` here. Maybe a treatment that shows a proper error in the output box instead?

John-Kryspin:
Yeah, that would work as well, but I thought it would be good to really blast this message into the user, as I worry if its in the output box the user might keep typing in more prices and see that message too late.

peter50216:
FYI, there's a similar PR at #349

theRTC204:
I appreciate your desire to contribute, but I have to agree with the other comments here. Use of an alert message is really not a good solution to this UX concern.

John-Kryspin:
No problem! I appreciate everyone’s input, I’ll close this PR

PR #388, opened 2020-05-20
Remove object spread operator so old edge works

### LINKED ISSUE #383: Output isn't showing any numbers when I put sales prices in...
**Things to check**
  - yes
  - yes
  - no

**Describe the bug**
When I put sales prices for turnips on each day for both AM and PM, the output numbers don't show up. It says output, predictions show nothing at all. I don't know why this is happening to me, but I hope you'll look into it.

**To Reproduce**

**Permalink to your prices**
https://turnipprophet.io?prices=101.109.93.104.110........

**Expected behavior**
I expected for the predictions to go there, but I see nothing.

**Screenshots**

 - OS: PC
 - Browser: Microsoft Edge
 - Version: Window's 10

**Smartphone (please complete the following information):**

**Additional context**

Lego6245:
Old edge doesn't support the spread operator for object literals. We can make a simple one line change that use Object.assign instead which does basically the same thing, but allows for users still using old edge to use the site.

We should consider a transpile step to fix this class of issues automatically for devs (and potentially show a banner warning for browsers that aren't supported)

Fixes #383

### DIFF
--- a/js/predictions.js
+++ b/js/predictions.js
@@ -321,7 +321,7 @@ class Predictor {
 
   * multiply_generator_probability(generator, probability) {
     for (const it of generator) {
-      yield {...it, probability: it.probability * probability};
+      yield Object.assign({probability: it.probability * probability}, it);
     }
   }
 


rctgamer3:
Is this broken in Edge (Webkit)?

PR #394, opened 2020-05-25
Add theme FR translation

Lysoun:


theRTC204 (approved):
Thank you for adding these.

PR #395, opened 2020-05-27
Add a High Color Contrast option for colorblind users

### LINKED ISSUE #254: Alternate color scheme for colorblind users

davidcorry:
This fixes #254, letting folks who are colorblind more easily differentiate chart colors. Some other tweaks were also included to further create contrast.

I was originally going to have it be a theme under the dropdown, but for simplicity's sake I just put it as a checkbox. This allows for automatic theme switching with this option selected. Should any other themes be added, they can easily add a high color contrast option as well.

I could potentially add it in as a dropdown option, but handling the logic for light/dark switching seemed tricky. If you want me to pursue it, though, let me know!


davidcorry:
Some previews of the light mode option:

image
image

Here's the dark mode option:

image

PR #397, opened 2020-05-28
add theming translations for chinese

binotaliu:


theRTC204 (approved):
Thank you for adding these 👍

PR #405, opened 2020-06-05
Filipino translation

valdotdev:
Did a Filipino (tagalog) translation since I have friends who prefer reading the text in Filipino. They can still cope with the english text since they _do_ play the game afterall but of course giving them options is the way to go.

Anyway I did my best in translating and here it is. There are some obvious terms that do not have true native Filipino translations. But in the Philippines, English is still used as a second language so it's pretty normal to add English jargon within Filipino statements ("tag-lish" so to speak) if needed especially with foreign terms.

Believe it or not a _lot_ of Filipinos don't even understand quite a lot of native Filipino words (since some are so "deep") and if I explicitly translate some of them here than I'm pretty sure they won't understand a big chunk of it.

### DIFF
--- a/js/translations.js
+++ b/js/translations.js
@@ -17,6 +17,7 @@ const LANGUAGES = {
   'ja': '日本語',
   'ko': '한국어',
   'nl': 'Nederlands',
+  'ph': 'Filipino',
   'pl': 'Polski',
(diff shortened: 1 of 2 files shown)


theRTC204 (approved):
Thank you for this contribution!

PR #412, opened 2020-06-20
Added Hebrew translation

RealA10N:
Please note that Hebrew is a right-to-left language. In order for the site to display the language well, something like `html { direction: rtl }` needs to be added to the HTML file only when the site is loaded in Hebrew.


theRTC204 (changes requested):
Thanks for the contribution.

We have a standard with the translations that the name Turnip Prophet is treated as a proper name, and thus not localized. Can you please revert all uses of Turnip Prophet to their English version?

In terms of the right-to-left issue you raised, I have no experience with RTL languages on the web. Would this translation work without the HTML attribute you mentioned? If not, we should probably work on an a mechanism for handling that attribute before merging this in.

RealA10N:
Hello there! Sorry for waiting so long 😅
I've reverted the translations of the "Turnip Prophet" string, as requested.

About the right-to-left issue: it is possible to merge and use the translation without the attribute, but some text (for example, sentences with both English and Hebrew text) will appear a bit messed up.

A really simple mechanism that will manage this attribute is needed indeed. Unfortunately, my knowledge in JS and HTML is very little and I'm afraid I can't do that. I would suggest saving the direction of each translation language somewhere in the locale translation JSON file and then loading it when the language is selected directly to the `direction` attribute.

The right-to-left issue is not related only to Hebrew. Would you like me to open a new GitHub issue about it?

PR #419, opened 2020-08-15
Add colour when hovering over the input, copy & reset button

JHSat:
Colour added when hovering over input, copy & reset button. It is same colour with those option button to reduce confusion.


theRTC204 (approved):

theRTC204:
Thanks for your contribution!

PR #422, opened 2020-10-02
Add subsource integrity, fix NL tl error

rctgamer3:
This adds subsource integrity to check that the referenced remote scripts have not been modified and fixes a NL translation error


mikebryant (approved):
Thank you :)

I'm sorry I've taken so long to look at this

PR #425, opened 2020-11-28
Update Hungarian translation

smileyhead:
Add error message at lines 87-89.

### DIFF
--- a/locales/hu.json
+++ b/locales/hu.json
@@ -84,4 +84,8 @@
       "dark": "Sötét"
     }
   }
+  "errors": {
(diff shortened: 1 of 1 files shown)


locales/hu.json
```diff
@@ -84,4 +84,8 @@
       "dark": "Sötét"
     }
   }
+  "errors": {
```
theRTC204, line 87: There appears to be a syntax error here. The closing brace immediately before this line needs to end with a comma in order to add an extra key to the object.

theRTC204 (changes requested):
Please make sure the JSON file is well formed.

rctgamer3 (approved):

